PR #2, opened 2021-04-05
Feature/upload experiment

mosheman5:
add option to upload the experiment directory to s3 bucket when the run is finished.

### DIFF
--- a/src/conf/experiment/defaults.yaml
+++ b/src/conf/experiment/defaults.yaml
@@ -3,4 +3,4 @@ experiment:
   debug: False
   manual_seed: 1234
   name: null
-  bucket_name: null
+  bucket_name: deepvoice-experiments
--- a/src/conf/experiment/defaults.yaml
+++ b/src/conf/experiment/defaults.yaml
@@ -3,4 +3,4 @@ experiment:
   debug: False
   manual_seed: 1234
   name: null
-  bucket_name: ???
+  bucket_name: null
--- a/src/conf/experiment/defaults.yaml
+++ b/src/conf/experiment/defaults.yaml
@@ -3,3 +3,4 @@ experiment:
   debug: False
   manual_seed: 1234
   name: null
+  bucket_name: ???
--- a/src/train.py
+++ b/src/train.py
@@ -22,11 +22,10 @@ import wandb
 from functools import partial
 from pathlib import Path
 from trainers import Trainer
-import logging
 import hydra
 from hydra.utils import instantiate
 import random
-from utils import Logger
+from utils import Logger, upload_experiment_to_s3
 from unittest.mock import Mock
 import os
 from utils import App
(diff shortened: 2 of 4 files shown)


src/utils.py
```diff
@@ -233,4 +235,43 @@ def __getattr__(self, item):
...
+        generator of files in directory tree
+    """
+    for p in Path(input_path).iterdir():
+        if p.is_dir():
+            yield from walk(p)
+            continue
+        yield p.resolve()
+
+
+def upload_experiment_to_s3(experiment_id: str,
```
noamzbr, line 256: I don't have any comments on this without running. Does it run?

Also, could you help us understand the upload structure? Currently in the outputs dir I have the following:
outputs/2021-03-04/22-43-10/
and in it I have the *.pth and wandb dir

What would that path translate to?
exp_id/2021-03-04/22-43-10/*
or without the date?

src/train.py
```diff
@@ -172,6 +171,10 @@ def main(args):
         model_args=args.model.model,
         logger=logger,
     )
+    if args.experiment.bucket_name and not args.experiment.debug:
```
noamzbr, line 174: Looks like this would run only on training loop completion. Personally most of my runs are terminated when I'm fed up with them :P So perhaps this should be (also?) an external python program? or perhaps we can catch the ctrl + c SIGINT like [here](https://www.devdungeon.com/content/python-catch-sigint-ctrl-c)

src/utils.py
mosheman5: with the date, since `include_parent=True` is default
I've tried the code, seems to work (both on dummy and real experiment folder)

src/train.py
mosheman5: external process probably would be the best, I wanted to have a basic option so we could easily upload existing experiments/catch the ones that trained til the end

src/conf/experiment/defaults.yaml
```diff
@@ -2,4 +2,5 @@
 experiment:
   debug: False
   manual_seed: 1234
-  name: null
\ No newline at end of file
+  name: null
+  bucket_name: null
```
Z30G0D, line 6: Why not put our s3://deepvoice-experiments/ bucket as ... well, defaults? :D
```suggestion
  bucket_name: 's3://deepvoice-experiments/ '
```

Z30G0D (commented):
Think this is great.
I saw this package on reddit the other day
https://www.reddit.com/r/MachineLearning/comments/mrcv7y/p_nimbo_run_jobs_on_aws_with_a_single_command/

https://docs.nimbo.sh/

Sounds really handy for us

src/conf/experiment/defaults.yaml
mosheman5: so it won't break for non-deepvoice users.

mosheman5:
> Think this is great.
> Related to this matter.
> I saw this package on reddit the other day
> https://www.reddit.com/r/MachineLearning/comments/mrcv7y/p_nimbo_run_jobs_on_aws_with_a_single_command/
> 
> https://docs.nimbo.sh/
> 
> Sounds really handy for us

Looks promising! we should give it a closer look

PR #4, opened 2021-04-19
fix inference

mosheman5:
enable inference on any model, specified by it's checkpoint args.
take also training sample rate and preprocessors to the inference args

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,9 +1,9 @@
-numba==0.48
+numba
 numpy
 torch
 torchaudio
 torchvision
-librosa==0.7.2
+librosa
 tqdm
 hydra-core
 pytest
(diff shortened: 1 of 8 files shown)


noamzbr (changes requested):
Looks good, can also remove sample_rate from inference_file.yaml

Also, can't we also pick seq_length from the loaded checkpoint?

amitgalor18 (commented):
Does the process of running inference using a checkpoint remain the same? Don't you need to specify the model architecture somewhere?

Or does the ckpt args contain everything needed, assuming the data has the right dimensions?

mosheman5:
> Looks good, can also remove sample_rate from inference_file.yaml
> 
> Also, can't we also pick seq_length from the loaded checkpoint?

We can and we should, updated


> Does the process of running inference using a checkpoint remain the same? Don't you need to specify the model architecture somewhere?
> 
> Or does the ckpt args contain everything needed, assuming the data has the right dimensions?

Exactly, the model is specified from the checkpoint args. Not sure I got the note about data dimension

PR #8, opened 2021-05-04
Feature/validate model confs

noamzbr:
Add features needed for google resnet-50 and for https://github.com/cchinchristopherj/Right-Whale-Convolutional-Neural-Network

### DIFF
--- a/src/conf/data/_preprocessors_sliding_window.yaml
+++ b/src/conf/data/_preprocessors_sliding_window.yaml
@@ -4,7 +4,11 @@ _preprocessors:
     _target_: torchaudio.transforms.Spectrogram
     n_fft: 1024
     hop_length: 256
+  sliding_window_norm:
+    _target_: data.SlidingWindowNormalize
+    n_fft: 1024
+    sr: ${data.sample_rate}
   amplitude_2_db:
     _target_: torchaudio.transforms.AmplitudeToDB
-  peak_norm:
-    _target_: data.PeakNormalize
+#  peak_norm:
+#    _target_: data.PeakNormalize
--- a/src/conf/data/_preprocessors_sliding_window.yaml
+++ b/src/conf/data/_preprocessors_sliding_window.yaml
@@ -6,9 +6,7 @@ _preprocessors:
     hop_length: 256
   sliding_window_norm:
     _target_: data.SlidingWindowNormalize
-    n_fft: 1024
+    n_fft: ${_preprocessors.spectrogram.n_fft}
     sr: ${data.sample_rate}
   amplitude_2_db:
     _target_: torchaudio.transforms.AmplitudeToDB
-#  peak_norm:
-#    _target_: data.PeakNormalize
--- a/src/data.py
+++ b/src/data.py
@@ -238,13 +238,13 @@ class SlidingWindowNormalize:
     """
 
     def __init__(self, sr: float, n_fft: int, lower_cutoff: float = 50, norm=True,
-                 inner: int = 3, outer: int = 32):
+                 inner_ratio: float = 0.06, outer_ratio: float = 0.5):
         self.sr = sr
         self.n_fft = n_fft
         self.lower_cutoff = lower_cutoff
         self.norm = norm
-        self.inner = inner
-        self.outer = outer
+        self.inner_ratio = inner_ratio
+        self.outer_ratio = outer_ratio
 
     # slidingWindowV Function from: https://github.com/nmkridler/moby2/blob/master/metrics.py
     def slidingWindowV(self, P):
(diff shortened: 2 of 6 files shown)


src/conf/data/_preprocessors_sliding_window.yaml
```diff
@@ -0,0 +1,14 @@
+# @package _global_
+_preprocessors:
+  spectrogram:
+    _target_: torchaudio.transforms.Spectrogram
+    n_fft: 1024
+    hop_length: 256
+  sliding_window_norm:
+    _target_: data.SlidingWindowNormalize
+    n_fft: 1024
```
mosheman5, line 9: if it needs to be aligned with the spectrogram you can use ${_preprocessors.spectrogram.n_fft}

src/data.py
```diff
@@ -227,6 +228,107 @@ def __call__(self, sample):
         return (sample - sample.mean()) / (sample.std() + 1e-8)
 
 
+class SlidingWindowNormalize:
+    """ Based on Sliding window augmentations of
+    https://github.com/cchinchristopherj/Right-Whale-Convolutional-Neural-Network/blob/master/whale_cnn.py
+        Translated to torch
+        Has 50/50 chance of activating H sliding window or V sliding window
+
+        Must come after spectrogram and AmplitudeToDB
```
mosheman5, line 237: I think it comes before AmplitudeToDB in the conf file

src/models.py
```diff
@@ -193,6 +193,19 @@ def forward(self, x):
         return out
 
 
+class GoogleResNet50withPCEN(PCENMixin, GoogleResNet50):
```
mosheman5, line 196: Not sure what's going on here. 
Double inheritance is a devil, do we really need it? can we just add self.pcen and apply it before the forward of GoogleResNet50? 
And I don't like the code of pcen, I think we should do a small refactor there (nothing to do with this pr though)

src/data.py
```diff
@@ -227,6 +228,107 @@ def __call__(self, sample):
...
+                    inner: int length of inner exclusion region (to calculate local mean)
+                    outer: int length of the window (to calculate overall mean)
+                    maxM: int size of the output image in the y-dimension
+                    norm: boolean indicating whether or not to cut off extreme values
+                Returns:
+                    Q: 2-D numpy array image with vertically-enhanced contrast
+
+        '''
+        min_f_ind = int((self.lower_cutoff / (self.sr / 2)) * self.n_fft)
+        Q = P.cpu().clone().numpy()
```
mosheman5, line 265: can we please change Q and P to q and p? or even better, let's give them indicative names :)

src/data.py
```diff
@@ -227,6 +228,107 @@ def __call__(self, sample):
...
+        wOuter = np.ones(int(self.outer_ratio * Q.shape[0]))
+        # Remove overall mean and local mean using np.convolve
+        for i in range(Q.shape[1]):
+            Q[:, i] = Q[:, i] - (np.convolve(Q[:, i], wOuter, 'same') - np.convolve(Q[:, i], wInner, 'same')) / (
+                        wOuter.shape[0] - wInner.shape[0])
+        Q[Q < 0] = 0.
+        return torch.from_numpy(Q).reshape(Q_shape)
+
+    # slidingWindowH Function from: https://github.com/nmkridler/moby2/blob/master/metrics.py
+    def slidingWindowH(self, P):
```
mosheman5, line 287: I think that most of the code here is similar to slidingWindowV, should we unite them to one function? create base function and two smaller ones?

mosheman5 (changes requested):
left some comments on the file diff

src/models.py
noamzbr: what's the problem? The PCENMixin is really elegant in my opinion, have a look at it. Basically it adds a (possibly trainable) PCEN layer before the original model in the forward method
About PCEN code - well, let's have a new PR

mosheman5: I don't have a problem with PCENMixin, I have a problem with the double inheritance, I don't think there is a justification to it.
the PCEN should be an attribute of GoogleResNet50withPCEN imo
let's zoom it

src/data.py
```diff
@@ -227,6 +228,84 @@ def __call__(self, sample):
...
+                        np.convolve(spect[i, :], wOuter, 'same') - np.convolve(spect[i, :], wInner, 'same')) / (
+                                      wOuter.shape[0] - wInner.shape[0])
+
+        spect[spect < 0] = 0.
+        return torch.from_numpy(spect).reshape(spect_shape)
+
+    def __call__(self, x):
+
+        if random.random() < 0.5:
+            return self.slidingWindow(x, dim=0)
```
mosheman5, line 304: Nice! :)

src/models.py
```diff
@@ -180,16 +184,18 @@ def forward(self, x):
...
 
-    def __init__(self, num_classes=2, eps=1E-6, s=0.025, alpha=0.98, delta=2, r=0.5, trainable=True):
-        super().__init__(num_classes=num_classes, eps=eps, s=s, alpha=alpha, delta=delta, r=r, trainable=trainable)
+    def __init__(self, num_classes=2, eps=1E-6, s=0.025, alpha=0.98, delta=2, r=0.5, trainable=False):
+        super().__init__(num_classes=num_classes)
+        self.pcen_model = PCENTransform(eps, s, alpha, delta, r, trainable)
 
     def forward(self, x):
-        out = super().forward(x)
+        out = self.pcen_model.forward(x)
```
mosheman5, line 197: I don't think you explicitly need to call forward here, since the __call__ method of nn.Module wraps the forward
simply `self.pcen_model(x)' will do

mosheman5 (approved):

PR #9, opened 2021-05-22
Feature/eda script

amitgalor18:
Added and formatted the content from EDA notebook to a function in utils.py. Simple for use, only needs the raven annotations path, recordings folder path and a name for the dataset (optional).

Now that I opened the pr I see some unwanted changes to inference.py. I think it's a change that was done to master that I didn't have in my branch for some reason, it will be fixed.


src/utils.py
```diff
@@ -275,3 +297,103 @@ def upload_experiment_to_s3(experiment_id: str,
...
+    Takes annotation files (selection table) created in Raven and turns it to a compatible annotations csv.
+    """
+    #create dataframe
+    annotations=Path(annotations_path)
+    recording=Path(recording_path)
+    filelist = list(annotations.glob('*selections.txt')) #ignore irrelevent files
+    metadata = []
+    for file in filelist:
+        dfTemp = pd.read_csv(file, sep="\t")
+        dfTemp['filename'] = re.search(r'18\d{4,4}_\d{6,6}', file.as_posix()).group()
```
noamzbr, line 316: Is it specific for our files? Because this looks specific

src/utils.py
```diff
@@ -275,3 +297,103 @@ def upload_experiment_to_s3(experiment_id: str,
...
+        dfTemp['StartMicInWater'] = np.amin(dfTemp['Begin Time (s)'])
```
noamzbr, line 317: Also this naming. Other datasets won't have mic going in and out of water so perhaps rename?

src/utils.py
```diff
@@ -275,3 +297,103 @@ def upload_experiment_to_s3(experiment_id: str,
...
+    for key, value in wav_filedict.items():
+        record_length = sf.info(value['path']).duration
+        value.update({'length': record_length})
+        wav_filedict[key] = value
+    annotation_lengths = []
+    for filename_txt in metadata['filename']:
+        annotation_lengths.append(wav_filedict[filename_txt]['length'])
+    metadata['TotalRecordLength'] = annotation_lengths
+    # filter metadata to contain only desired call types
+    filters = ['w', 'sc']
```
noamzbr, line 338: Again, if we aim to be generic to raven files, need to remove stuff specific to our annotations...

src/utils.py
```diff
@@ -275,3 +297,103 @@ def upload_experiment_to_s3(experiment_id: str,
     for upload_file in tqdm(upload_files):
         upload_file = str(upload_file)
         s3_client.upload_file(upload_file, bucket_name, upload_file.replace(current_global, object_global))
+
+
+def raven_to_csv_annotations(annotations_path: str,
+                             recording_path: str,
+                             dataset_name: str = 'recordings_2018_filtered'):
```
noamzbr, line 304: Perhaps generate random name like wandb?

src/utils.py
```diff
@@ -275,3 +297,103 @@ def upload_experiment_to_s3(experiment_id: str,
...
+    bg_segments = pd.DataFrame(bg_segments, columns=['begin_time', 'end_time', 'filename'])
+    bg_segments['call_length'] = bg_segments['end_time'] - bg_segments['begin_time']
+    bg_segments.sort_values(by=['call_length'])
+    # add labels: 0 for background and 1 for call. TODO: modify if there are different call types
+    bg_segments['label'] = np.zeros(bg_segments.shape[0], dtype=int)
+    non_overlap_calls['label'] = np.ones(non_overlap_calls.shape[0], dtype=int)
+    # combine to a csv
+    combined_annotations = pd.concat([bg_segments, non_overlap_calls])
+    filename = 'combined_annotations_'+dataset_name+'.csv'
+    combined_annotations.to_csv(filename, index=False)
```
noamzbr, line 365: This may be petty, but it would be prettier to have this function return the df and a separate function call it and save to csv

src/utils.py
amitgalor18: you're right, I'll generalize it

src/utils.py
amitgalor18: This seems less relevant since it doesn't affect the output. I removed all the actual code regarding out-of-water sections

src/utils.py
amitgalor18: as discussed, I'll add it as an input

src/utils.py
amitgalor18: The user can enter a name if they have a different dataset, but I think that since we publish it with our dataset as an example it can be simple to have the default as our dataset name. 
@mosheman5  what do you think?

src/inference.py
```diff
@@ -74,26 +74,29 @@ def predict(model: torch.nn.Module, data_loader: Generator[torch.tensor, None, N
         return (predicted_probability > threshold).reshape((-1, 1))
 
 
-def load_model(path):
+def load_model(model_params, checkpoint_state_dict):
```
mosheman5, line 77: please use the updated version of master

src/utils.py
mosheman5: Either remove or keep the default, I see no use in using random name for dataset (the frequency of creating a dataset is much lower than the frequency of running an experiment)

src/utils.py
mosheman5: so remove this field

src/utils.py
mosheman5: I also won't work for 2019 tags. 
I think that maybe separating each row by tabs/spaces will return the needed fields

src/utils.py
```diff
@@ -275,3 +297,103 @@ def upload_experiment_to_s3(experiment_id: str,
...
+        dfTemp = pd.read_csv(file, sep="\t")
+        dfTemp['filename'] = re.search(r'18\d{4,4}_\d{6,6}', file.as_posix()).group()
+        dfTemp['StartMicInWater'] = np.amin(dfTemp['Begin Time (s)'])
+        dfTemp['EndMicInWater'] = np.amax(dfTemp['End Time (s)'])
+        metadata.append(dfTemp)
+
+    metadata = pd.concat(metadata)
+    metadata.rename(columns={'Begin Time (s)': 'begin_time', 'End Time (s)': 'end_time'}, inplace=True)
+    print('Number of Labels:', metadata.shape[0])
+    metadata['Annotation'] = metadata['Annotation'].replace(np.nan, 'w', regex=True)
```
mosheman5, line 324: maybe we should define positive call indicator as and argument for the method

src/utils.py
```diff
@@ -275,3 +297,103 @@ def upload_experiment_to_s3(experiment_id: str,
...
+    #add recording length to dataframe
+    wav_filelist = list(recording.glob('*.wav'))
+    wav_filedict = {re.search(r'18\d{4,4}_\d{6,6}', file.as_posix()).group(): {'path': file} for file in wav_filelist
```
mosheman5, line 327: again, too specific

src/utils.py
```diff
@@ -275,3 +297,103 @@ def upload_experiment_to_s3(experiment_id: str,
...
+    bg_segments = []
+    for file in (unique_files):
+        file_df = non_overlap_all[non_overlap_all['filename'] == file]
+        begin = np.array(file_df['begin_time'])
+        end = np.array(file_df['end_time'])
+        for item in end:
+            next_beginning = begin[begin > item]
+            if next_beginning.size == 0:
+                break
+            next_beginning = np.min(next_beginning)
```
mosheman5, line 354: next_beginning = next_beginning[0] ? I don't remember if the array is order but I believe it is

src/utils.py
amitgalor18: worked on it today. changed the search to ".selection" so it's generalized to any raven file.
I thought about having it look for ".txt" to be even more generalized, and then instruct the user to not have any other text files in the folder, but we first need to do it ourselves (i.e. we have txt files in our finding_willy/annotations, so it wouldn't have worked for you guys had done it).

src/inference.py
amitgalor18: Merged master into the branch today. Says it's up to date.

src/utils.py
amitgalor18: I just noticed this comment. I didn't understand what you mean.

amitgalor18:
Other than Moshe's last comment, I think we can close this pr. let me know if you approve.

Z30G0D:
@amitgalor18 very good!
There was a conflict somehow with master.
please check :
https://github.com/deep-voice/soundbay/pull/9/commits/3a4ce7a0fa5582ce7bc324517351a855b091e776
To see if I solved it ok

amitgalor18:
@Z30G0D I don't know how it happened since I merged master into the branch when I commented a week ago. But your merge looks good to me.

PR #10, opened 2021-06-21
fix chrisop to match repo

noamzbr:
Fix chrisop model to match 
https://github.com/cchinchristopherj/Right-Whale-Convolutional-Neural-Network/blob/master/training.py

### DIFF
--- a/src/conf/optim/christoph.yaml
+++ b/src/conf/optim/christoph.yaml
@@ -6,6 +6,3 @@ optim:
     lr: 0.1
     weight_decay: 0.005
     nesterov: False
-  scheduler:
-    _target_: torch.optim.lr_scheduler.ExponentialLR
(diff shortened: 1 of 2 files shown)


src/conf/optim/christoph.yaml
```diff
@@ -0,0 +1,11 @@
...
+optim:
+  epochs: 1000
+  optimizer:
+    _target_: torch.optim.SGD
+    lr: 0.1
+    weight_decay: 0.005
+    nesterov: False
+  scheduler:
+    _target_: torch.optim.lr_scheduler.ExponentialLR
+    gamma: 1
```
mosheman5, line 11: do we need a scheduler if gamma is 1?

mosheman5 (approved):
lgtm

PR #11, opened 2021-06-21
Visualization and stuff

Z30G0D:
Figure_1

Need to handle the y axes in a better way, WIP


src/train.py
```diff
@@ -119,8 +119,10 @@ def main(args):
 
     # Set device
     if not torch.cuda.is_available():
+        print('CPU!!!!!!!!!!!')
```
mosheman5, line 122: lol

src/conf/data/defaults.yaml
```diff
@@ -6,13 +6,13 @@ data:
   data_sample_rate: 44100
   train_dataset:
     _target_: data.ClassifierDataset
-    data_path: '../tests/assets/data/'
+    data_path: '../datasets/recordings_2018/'
```
mosheman5, line 9: Do we want to change the default file? I think we should have another configuration for running the deep voice dataset.

src/conf/data/defaults.yaml
```diff
@@ -6,13 +6,13 @@ data:
...
     mode: train
-    metadata_path: '../tests/assets/annotations/sample_annotations.csv'
+    metadata_path: '../annotations/combined_annotations_filtered_train.csv'
     augmentations_p: 0.8
     augmentations: ${_augmentations}
     preprocessors: ${_preprocessors}
-    seq_length: 1
+    seq_length: 0.3
```
mosheman5, line 15: why?

src/Notebooks/recording_vIsualization.ipynb
```diff
@@ -0,0 +1,374 @@
+{
```
noamzbr, line 1: Perhaps worth separating the function from the notebook?

src/Notebooks/recording_vIsualization.ipynb
```diff
@@ -0,0 +1,374 @@
...
+ "cells": [
```
noamzbr, line 2: Should add start_time & end_time so it won't always show all of the file.

src/Notebooks/recording_vIsualization.ipynb
```diff
@@ -0,0 +1,374 @@
...
+  {
```
noamzbr, line 3: Add defaults to thresh, seg_length in function header

src/Notebooks/recording_vIsualization.ipynb
```diff
@@ -0,0 +1,374 @@
...
+   "cell_type": "code",
```
noamzbr, line 4: Optional - if to many segments, separate into a few plots so plot won't be cramped

noamzbr (changes requested):
Real nice! Added a few comments / small feature requests

src/Notebooks/recording_vIsualization.ipynb
```diff
@@ -0,0 +1,374 @@
...
+   "execution_count": 159,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# idx= 10\n",
+    "# warnings.filterwarnings('ignore')\n",
+    "sample = xticks\n",
+    "%matplotlib qt\n",
```
mosheman5, line 287: should add PyQt5 to the requirements.txt file

src/Notebooks/recording_vIsualization.ipynb
```diff
@@ -0,0 +1,374 @@
...
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Nu mber of tags to inspect: 488\n"
+     ]
+    }
+   ],
+   "source": [
+    "print('Number of seconds:', tags_to_inspect.shape[0])"
```
mosheman5, line 168: tags_to_inspect is not defined before. should be removed?

src/Notebooks/recording_vIsualization.ipynb
```diff
@@ -0,0 +1,374 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 157,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def play_audio_plot_stft(prob, df_series, audio_path, thresh, seg_length):\n",
+    "    def get_sr(wav_path):\n",
```
mosheman5, line 192: you can use soundfile.info(wav_path).samplerate instead

src/Notebooks/recording_vIsualization.ipynb
```diff
@@ -0,0 +1,374 @@
...
+    "        with wave.open(str(wav_path), \"rb\") as wave_file:\n",
+    "            return wave_file.getframerate()\n",
+    "        \n",
+    "    \n",
+    "    def play_audio(audio_data, sr):\n",
```
mosheman5, line 197: do we use it?

mosheman5 (changes requested):
IMO, plotting all the graphs on one fig is bit "heavy" and I can't really see the details in the spectrogram.
I suggest breaking it down a plot for each graph, and to add an option to play the audio of this graph. 
If you'll use `%matplotlib notebook` instead of `%matplotlib qt` you'll have all the graphs inside the notebook, instead of a popup windows (having 10 popup windows isn't nice)

src/Notebooks/recording_vIsualization.ipynb
Z30G0D: maybe not, just for convenience changing the colors easily

PR #13, opened 2021-06-27
add ICML 2013 data conf and annotations notebook

noamzbr:


mosheman5 (approved):
Didn't inspect the notebook, tell me if you think it's needed

Z30G0D (approved):
Hey,
changed the notebook a bit (maybe it's a good time to use Moshe's suggestion for diff of notebooks :) )
Main change is :
`
annotation_df = annotation_df.astype({'begin_time': 'float64', 'end_time':'float64', 'call_length':'float64',
                                     'label':'int32'})`

The dataframe is creating as a default an object type instead of str (for file name) or floats (for begin/ end times..)

I guess it could be a difference in pandas version, but might as well leave it for clarity.

Wonderful work!

PR #15, opened 2021-07-12
fix model and preprocessing to match repo

noamzbr:


amitgalor18 (commented):
The changes to the model and optimizer of this specific christopher model look good.
But is the change to the preprocessors a global change? I mean, if it changes the default fft window size we need to make sure it doesn't ruin the training of different models.
Or is the "_preprocessors_sliding_window" file only used by the christoph model?

src/models.py
```diff
@@ -100,18 +100,20 @@ def __init__(self, num_classes=2):
...
+        self.fc1 = nn.Linear(11 * 11 * 30, 200)
+        self.fc2 = nn.Linear(200, num_classes)
         self.drop_out = nn.Dropout(0.5)
 
     def forward(self, x):
         out = self.drop_out(x)
         out = self.layer1(out)
         out = self.layer2(out)
-        out = self.avgpool(out)
+        # out = self.avgpool(out)
```
mosheman5, line 112: better do delete the module from init and here instead of comment out

noamzbr:
> The changes to the model and optimizer of this specific christopher model look good.
> But is the change to the preprocessors a global change? I mean, if it changes the default fft window size we need to make sure it doesn't ruin the training of different models.
> Or is the "_preprocessors_sliding_window" file only used by the christoph model?

Basically yeah. Others may use it but it makes sense to have there the settings for the model that currently uses that.

noamzbr:
@mosheman5 permission to merge?

src/models.py
```diff
@@ -97,21 +97,20 @@ def __init__(self, num_classes=2):
...
 
     def forward(self, x):
         out = self.drop_out(x)
         out = self.layer1(out)
         out = self.layer2(out)
-        out = self.avgpool(out)
         out = torch.flatten(out, 1)
-        out = self.fc2(out)
+        out = self.fc1(out)
+        out = torch.nn.ReLU()(out)
```
mosheman5, line 111: More elegant to configure in the init 
`self.relu = torch.nn.ReLU()` and use here `out = self.relu(out)`
Another option is to use the functional and pass the object creation on each forward pass:
`out = torch.nn.functional.relu(out)`

mosheman5:
> @mosheman5 permission to merge?

Yalla

PR #16, opened 2021-07-12
add notebook and data yaml for audio MNIST

noamzbr:
Please let me know if you want me to add also the data itself + annotation to our repo. It takes around 30Mb.

If not, our users will need to follow the notebook to the git and produce their own data and annotations.


amitgalor18:
To my (not so extensive) knowledge, it's more common to have a link to the dataset (on a website or google drive) than to upload it directly to the repo. As I understand our users will mostly use this as a service and less as a standalone product so there's also the option to have it ready to be shared upon request by mail.

Z30G0D:
I think it's safe to upload it to s3.

PR #17, opened 2021-07-17
flatten dict for wandb

mosheman5:
Flatten the args before logging them to wandb.
PR by @AsafMazuz1

### DIFF
--- a/src/train.py
+++ b/src/train.py
@@ -113,7 +113,7 @@ def main(args):
 
     # Set logger
     # if not args.experiment.debug else Mock()
-    _logger = wandb
+    _logger = wandb if not args.experiment.debug else Mock()
     _logger.init(project="finding_willy", name=args.experiment.name)
 
     # Set device
--- a/src/train.py
+++ b/src/train.py
@@ -114,7 +114,7 @@ def main(args):
     # Set logger
     # if not args.experiment.debug else Mock()
     _logger = wandb
-    _logger.init(project="testing", name=args.experiment.name)
+    _logger.init(project="finding_willy", name=args.experiment.name)
 
     # Set device
     if not torch.cuda.is_available():
@@ -138,8 +138,8 @@ def main(args):
 
     # Logging
     logger = Logger(_logger)
-    args = flatten(args)
-    logger.log_writer.config.update(args)
+    flattenArgs = flatten(args)
+    logger.log_writer.config.update(flattenArgs)
     App.init(args)
 
     # Define criterion
--- a/src/train.py
+++ b/src/train.py
@@ -25,11 +25,10 @@ from trainers import Trainer
 import hydra
 from hydra.utils import instantiate
 import random
-from utils import Logger, upload_experiment_to_s3
+from utils import Logger, upload_experiment_to_s3, flatten
 from unittest.mock import Mock
 import os
 from utils import App
-
 
 
 def modeling(
@@ -44,7 +43,6 @@ def modeling(
     model_args,
     logger,
 ):
-
     """
     modeling function takes all the variables and parameters defined in the main script
     (either through hydra configuration files or overwritten in the command line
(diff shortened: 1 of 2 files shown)


src/train.py
```diff
@@ -109,12 +107,14 @@ def modeling(
...
+# TODO check how to use hydra without path override
+@hydra.main(config_name="runs/main", config_path="conf")
 def main(args):
 
     # Set logger
-    _logger = wandb if not args.experiment.debug else Mock()
-    _logger.init(project="finding_willy", name=args.experiment.name)
+    # if not args.experiment.debug else Mock()
+    _logger = wandb
+    _logger.init(project="testing", name=args.experiment.name)
```
mosheman5, line 117: rename the project to 'finding_willy'

src/train.py
```diff
@@ -138,6 +138,7 @@ def main(args):
 
     # Logging
     logger = Logger(_logger)
+    args = flatten(args)
```
mosheman5, line 141: Wouldn't override `args` since we need it in other parts in the code. Would rename and create new variable

src/utils.py
```diff
@@ -13,6 +13,13 @@
 import pandas as pd
 import re
 
+import collections
+
+try:
+    collectionsAbc = collections.abc
+except AttributeError:
+    collectionsAbc = collections
```
mosheman5, line 21: what's the use case for the try n except here?

src/utils.py
```diff
@@ -440,3 +473,14 @@ def merge_with_checkpoint(run_args, checkpoint_args):
     run_args.data.sample_rate = checkpoint_args.data.sample_rate
     OmegaConf.set_struct(run_args, True)
     return run_args
+
+
+def flatten(d, parent_key='', sep='_'):
```
mosheman5, line 478: please use `sep='.'` instead of `sep='_'`

src/utils.py
```diff
@@ -102,17 +111,18 @@ def __init__(self,
         self.pred_list = []
         self.pred_proba_list = []
         self.label_list = []
-        self.metrics_dict = {'accuracy': [], 'f1score': [], 'precision': [], 'recall': [], 'auc': []}
+        self.metrics_dict = {'accuracy': [], 'f1score': [],
+                             'precision': [], 'recall': [], 'auc': []}
```
mosheman5, line 115: Can you change your max line length in the code editor? it create a lot of diffs and it's hard to follow..

src/utils.py
AsafMazuz1: Fallback version for this library.
Depends on the python version

src/utils.py
AsafMazuz1: Set the value to?

Z30G0D: 120 characters is default seems to be sufficient since most of us use IntelliJ's Pycharm

src/utils.py
```diff
@@ -440,3 +473,14 @@ def merge_with_checkpoint(run_args, checkpoint_args):
     run_args.data.sample_rate = checkpoint_args.data.sample_rate
     OmegaConf.set_struct(run_args, True)
     return run_args
+
+
+def flatten(d, parent_key='', sep='.'):
```
Z30G0D, line 478: Do we also need a reverse method (unflatten) to turn back to yaml?
Just thinking out loud

Z30G0D (commented):
Very nice job!
Small comment.

src/train.py
```diff
@@ -109,11 +107,13 @@ def modeling(
...
 
-@hydra.main(config_name="runs/main", config_path="conf")  # TODO check how to use hydra without path override
+# TODO check how to use hydra without path override
+@hydra.main(config_name="runs/main", config_path="conf")
 def main(args):
 
     # Set logger
-    _logger = wandb if not args.experiment.debug else Mock()
+    # if not args.experiment.debug else Mock()
+    _logger = wandb
```
noamzbr, line 116: why remove that and log debug runs?

src/utils.py
```diff
@@ -440,3 +473,14 @@ def merge_with_checkpoint(run_args, checkpoint_args):
...
+    items = []
+    for k, v in d.items():
+        new_key = parent_key + sep + k if parent_key else k
+        if isinstance(v, collectionsAbc.MutableMapping):
+            items.extend(flatten(v, new_key, sep=sep).items())
```
noamzbr, line 483: Think `items.update(flatten(v, new_key, sep=sep))` is more readable but really not important

src/utils.py
noamzbr: I tend to agree, useful for actually running with past configurations. 
Among other things it would mean that we can't use . as a default separator. Note sure what we can use because nearly other char has a use in python (except , which is not elegant). Ideas?

noamzbr (changes requested):
All in all looks simple and would do the work. I think an inverse operation is must (as @Z30G0D said) and that it would be better to remove the diffs originating in IDE settings on line length.

src/train.py
AsafMazuz1: This was for my testing.
I will fix this in my next push :)

noamzbr (approved):

src/utils.py
noamzbr: @mosheman5 convinced me this is not needed
What we do need is a more fool proof approach to uploading experiments to s3

PR #20, opened 2021-08-14
Feature/sweeps

noamzbr:
Working sweep for ICML 2013

Didn't find a way with current repo to specify that a certain model should go with a certain optimizer. Current seep will try various combinations. Way to do this would be to create a meta config for both that specifies inside what model and optim are needed.


src/conf/augmentations/_augmentations.yaml
```diff
@@ -14,7 +14,7 @@ _augmentations:
     _target_: data_augmentation.FrequencyMasking
     p: 0.5
     sr: ${data.sample_rate}
-    min_freq_in_segment: 500
-    max_freq_in_segment: 1500
+    min_freq_in_segment: 200
+    max_freq_in_segment: ${data.max_freq}
```
mosheman5, line 18: Nice bugfix!

src/conf/data/audio_MNIST.yaml
```diff
@@ -4,6 +4,9 @@ data:
   num_workers: 10
   sample_rate: 8000
   data_sample_rate: 8000
+  max_freq: 4000
+  n_fft: 256
+  hop_length: 64
   train_dataset:
     _target_: data.ClassifierDataset
     data_path: 'PATH_TO_AUDIO_MNIST_DATA'
```
mosheman5, line 12: I know it's not part of the pr, but what?

src/train.py
```diff
@@ -113,7 +113,15 @@ def main(args):
 
     # Set logger
     _logger = wandb if not args.experiment.debug else Mock()
-    _logger.init(project="finding_willy", name=args.experiment.name)
+    if args.experiment.name:
```
mosheman5, line 116: Maybe wrap the logging handling into function under utils? Too much code to be on main, imo
btw, do you think we should have utils folder and seperate the functions to different files with relavant grouping?

src/utils.py
```diff
@@ -434,33 +434,12 @@ def non_overlap_df(input_df):
...
+    items = {}
     for k, v in d.items():
         new_key = parent_key + sep + k if parent_key else k
         if isinstance(v, collectionsAbc.MutableMapping):
-            #items.extend(flatten(v, new_key, sep=sep).items())
             items.update(flatten(v, new_key, sep=sep))
         else:
-            items.append((new_key, v))
+            items.update({new_key: v})
     return dict(items)
```
mosheman5, line 445: If i's already a dict, do we need to change the type here?

src/utils.py
```diff
@@ -434,33 +434,12 @@ def non_overlap_df(input_df):
     return non_overlap
 
 
-def merge_with_checkpoint(run_args, checkpoint_args):
```
mosheman5, line 437: Why? Don't we need it?

src/utils.py
noamzbr: it's a dict so it needs to use update and have a dict as argument. append.(tuple) is for a list of tuples.

src/utils.py
noamzbr: It's a duplicate

src/conf/data/audio_MNIST.yaml
noamzbr: I changed the yaml structure such that these parameters are in data and not in preprocessing. Makes way more sense.

mosheman5: I meant this:
`data_path: 'PATH_TO_AUDIO_MNIST_DATA'`
:)

mosheman5 (approved):

src/conf/data/audio_MNIST.yaml
noamzbr: oh, on purpose I don't have paths to data, cause it can by anywhere on user machine

src/train.py
noamzbr: Think that stuff about logging metrics should move to it's own place

tests/test_pipeline.py
```diff
@@ -38,6 +41,7 @@ def check_variable_change(model_before, model_after, vars_change=True, device=to
 
 def test_trainer(model, optimizer, scheduler, data_loader, criterion):
 
+    wandb.init(project=None, mode='disabled')
```
noamzbr, line 44: @mosheman5  Found an easier way to fix the test so undone some of your stuff

PR #22, opened 2021-12-04
add setup.py and minor additions.

noamzbr:


src/model_data_sweep.yaml
```diff
@@ -0,0 +1,26 @@
...
+    value: 100
+  data:
+    values:
+      - defaults
+      - icml2013
+      - orcasound
+  preprocessors:
+    value: _preprocessors_sliding_window
+  data.train_dataset.augmentations_p:
+    value: 0
```
mosheman5, line 18: There are no augmentatios on christophs model? weird

setup.py
```diff
@@ -0,0 +1,38 @@
...
+from distutils.util import convert_path
+import os
+
+main_ns = {}
+ver_path = convert_path('src/version.py')
+with open(ver_path) as ver_file:
+    exec(ver_file.read(), main_ns)
+VER = main_ns['__version__']
+
+requirementPath = os.path.dirname(os.path.realpath(__file__)) + '/requirements.txt'
```
mosheman5, line 12: is there a reason for not working with relative paths?

setup.py
```diff
@@ -0,0 +1,38 @@
...
+
+setup(
+    name='soundbay',
+    version=VER,
+    packages=setuptools.find_packages(),
+    install_requires=install_requires,
+    license_files='LICENCE',
+    description='Deep Learning Framework for Bioacoustics',
+    author='soundbay',
+    author_email='deepvoicengo@gmail.com',
```
mosheman5, line 26: change to info@deepvoicefoundation.com please

setup.py
```diff
@@ -0,0 +1,38 @@
...
+    exec(ver_file.read(), main_ns)
+VER = main_ns['__version__']
+
+requirementPath = os.path.dirname(os.path.realpath(__file__)) + '/requirements.txt'
+install_requires = []
+if os.path.isfile(requirementPath):
+    with open(requirementPath) as f:
+        install_requires = f.read().splitlines()
+
+setup(
```
mosheman5, line 18: we should also include torch and it's minions in the installation, or at least validate their versions

mosheman5 (changes requested):
Only the torch issue bothers me, other than that looks awesome!

setup.py
mosheman5: I think it might be a problem, since installing torch depends on the install cuda toolkit version. So guess we should either check this or demand torch & friends as a prerequisite

setup.py
```diff
@@ -0,0 +1,38 @@
...
+    name='soundbay',
+    version=VER,
+    packages=setuptools.find_packages(),
+    install_requires=install_requires,
```
mosheman5, line 22: perhaps we should add version to every package we use? Might be safe, so we won't end up with errors in dependency

setup.py
noamzbr: Mmm I really don't know how to do this, can you look at the requirments of other torch dependant libraries? I tried to look here but honestly didn't understand what's going on
https://github.com/pytorch/captum/blob/master/setup.py

setup.py
noamzbr: Would try to add <> condtions, we don't want setup to be restrictive

setup.py
noamzbr: Saw it's common in other repos. Safer this way I guess.

src/model_data_sweep.yaml
noamzbr: I seem to remember that was the winner in the seep, but not sure.

noamzbr:
ToDo:
1. Add note in readme that we require python and pytorch preinstalled
2. Fix issues with importing sub packages from soundbay

setup.py
mosheman5: As we discussed on slack, we can just a pytorch version as a requirement and the user will install it manually

mosheman5 (approved):
LGTM

requirements.txt
```diff
@@ -1,17 +1,17 @@
-numba
-numpy
-librosa
-tqdm
+numba>=0.54.0
+numpy<1.21.0
```
mosheman5, line 2: why `numpy<1.21.0`? and all the rest ar >=

noamzbr: numba has this requirment

noamzbr: numba requiremnt

PR #24, opened 2022-01-08
add ICML 2013 annotation

noamzbr:


soundbay/conf/runs/main.yaml
```diff
@@ -2,7 +2,7 @@
 defaults:
   - data: defaults
   - augmentations: _augmentations
-  - preprocessors: _preprocessors
+  - preprocessors: _preprocessors_unit
```
mosheman5, line 5: Did we decide unit normalization to occur on default?

soundbay/conf/data/icml2013.yaml
```diff
@@ -26,7 +26,7 @@ data:
     _target_: soundbay.data.ClassifierDataset
     data_path: 'PATH_TO_ICML2013_WAV_FILES'
     mode: val
-    metadata_path: 'PATH_TO_VAL_ANNOTATIONS'
+    metadata_path: '../annotations/ICML_2013_annotations/annotation_train2_val.csv'
```
mosheman5, line 29: why train2 and not just train? :P

mosheman5:
I think it's better to have a datasets hierarchy and save the annotations under the relevant dataset, as in here:
https://github.com/deep-voice/soundbay/tree/master/datasets
wdyt? 
I think it would be good to move the current files from annotations to `datasets/mozambiqe_2018` and remove annotations folder

mosheman5 (approved):
Thanks for the changes, looks great!
Can you add a word about where we can find the processed data, so ppl could reproduce the experiment?

PR #28, opened 2022-02-08
accommodate non-integer seq_length by changing dataloader read input type

amitgalor18:
moved int conversion from data reading in data.py to the definition of beginning times in getitem

### DIFF
--- a/soundbay/data.py
+++ b/soundbay/data.py
@@ -38,6 +38,8 @@ class ClassifierDataset(Dataset):
         self.audio_dict = self._create_audio_dict(Path(data_path))
         self.metadata_path = metadata_path
         metadata = pd.read_csv(self.metadata_path)
+        metadata_filenames = pd.read_csv(self.metadata_path, usecols=['filename'], dtype=pd.StringDtype())
+        metadata['filename'] = metadata_filenames
         self.metadata = self._update_metadata_by_mode(metadata, mode, split_metadata_by_label)
         self.mode = mode
         self.seq_length = seq_length
@@ -358,7 +360,7 @@ class InferenceDataset(Dataset):
             audio_dict contains references to audio paths given name from metadata
         """
         audio_len = sf.info(self.file_path).duration
-        self._start_times = np.arange(0, audio_len//self.seq_length * self.seq_length, self.seq_length).astype(int)
+        self._start_times = np.arange(0, audio_len//self.seq_length * self.seq_length, self.seq_length)
 
     def _get_audio(self, begin_time):
         """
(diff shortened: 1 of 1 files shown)


soundbay/data.py
```diff
@@ -38,6 +38,8 @@ def __init__(self, data_path, metadata_path, augmentations, augmentations_p, pre
         self.audio_dict = self._create_audio_dict(Path(data_path))
         self.metadata_path = metadata_path
         metadata = pd.read_csv(self.metadata_path)
+        metadata_filenames = pd.read_csv(self.metadata_path, usecols=['filename'], dtype=pd.StringDtype())
```
mosheman5, line 41: what  `metadata_filenames` is used for?

amitgalor18:
Oh, it's a fix to another bug I had with the filenames in the metadata csv.
I read this one column with the "StringDtype" and then replaced it in the
full csv, so it won't read it as numbers and shorten the filenames in the
resulting dataframe. The names are supposed to be strings anyways, so I
think the fix is relevant for master and not just the manatee branch.

Z30G0D (changes requested):
maybe change `dtypes` to a dictionary to avoid the "double" line.
And it looks great.

soundbay/data.py
amitgalor18: would that mean to dictate the type of all the possible columns in the metadata (which could be different with different datasets) or can I define a dict such as `dtype={'filename': 'str'}` ?
Upon writing the comment it makes sense that it's the latter, but I want to make sure

soundbay/data.py
```diff
@@ -358,7 +360,7 @@ def _create_start_times(self):
             audio_dict contains references to audio paths given name from metadata
         """
         audio_len = sf.info(self.file_path).duration
-        self._start_times = np.arange(0, audio_len//self.seq_length * self.seq_length, self.seq_length).astype(int)
+        self._start_times = np.arange(0, audio_len//self.seq_length * self.seq_length, self.seq_length)
```
noamzbr, line 363: doesn't it make sense to change this to / rather than // ?

soundbay/data.py
tomer-grin: Yes,
Create this dict in the `__init__` method of the dataloader and then call it in the `dtypes` argument when we call the CSV reader.
Something like this:

 `pd.read_csv(...,dtypes=self.metadata_columns_type)`

soundbay/data.py
amitgalor18: I think it would make the array contain a start time of a segment that's distance from the end of the wav is shorter than seq_length.
e.g.:
len=audio_len = 10.1 sec, and seq=seq_length=0.2 would give us:
np.arange(0,len//seq*seq,seq) = 0, 0.2, 0.4, ..., 9.6, 9.8
np.arange(0,len/seq*seq,seq) = 0, 0.2, 0.4, ..., 9.6, 9.8, 10
so the sequence starting at 10 with seq length 0.2 would exceed the wav length

mosheman5 (approved):
Good catch, merge plz

soundbay/data.py
noamzbr: oh understood

Z30G0D (approved):

PR #30, opened 2022-02-19
Add example of model exploration notebook

noamzbr:
Added example of notebook that loads a trained model and explores it's results on the validation dataset.


mosheman5 (approved):
Nice notebook, might be nice to merge all the model setup to one cell

PR #31, opened 2022-03-06
Add squeezenet model and conf

noamzbr:


amitgalor18 (approved):

soundbay/models.py
```diff
@@ -27,6 +30,114 @@ def _choose_block_class(block):
         return getattr(importlib.import_module(module_name), class_name)
 
 
+class SqueezeNet1D(nn.Module):
```
mosheman5, line 33: is there a reason to copy the model instead of importing it from torch?
And if there is a change from the torch version, could you please specify it?

mosheman5 (changes requested):
Big like for the tests update :)
I wonder about the code of the model, I'm not a fan of importing identical code from a package we already use. 
If we do we should explicitly state the source it was taken from/reason it.
Also, perhaps we should have `extern` folder for models we take from other places?

soundbay/models.py
noamzbr: the change was the 1 input channel in the first Conv2d, but updated to a version that uses the original definition.

mosheman5 (approved):

PR #33, opened 2022-03-08
added inference_csv_to_raven function 

amitgalor18:
added a conversion to raven format. Currently utilized it in the inference to file function (so it saves both a probability csv and a binary raven csv with some threshold. 
Also added conf files for eric manatee data inference.

### DIFF
--- a/soundbay/inference.py
+++ b/soundbay/inference.py
@@ -134,9 +134,12 @@ def inference_to_file(
     output_file = output_path / filename
     concat_dataset.to_csv(index=False, path_or_buf=output_file)
 
-    raven_filename = f"raven_annotations-{datetime.datetime.now().strftime('%Y-%m-%d_%H-%M-%S')}-{model_name}-{dataset_name}-thresh03.csv"
+    # save raven file
+    threshold = 0.5
+    thresholdtext = int(threshold*10)
+    raven_filename = f"raven_annotations-{datetime.datetime.now().strftime('%Y-%m-%d_%H-%M-%S')}-{model_name}-{dataset_name}-thresh0{thresholdtext}.csv"
     raven_output_file = output_path / raven_filename
-    raven_df = inference_csv_to_raven(results_df, test_dataset, threshold=0.5)
+    raven_df = inference_csv_to_raven(results_df, test_dataset, threshold=threshold)
     raven_df.to_csv(index=False, path_or_buf=raven_output_file, sep="\t")
 
     return
--- a/soundbay/inference.py
+++ b/soundbay/inference.py
@@ -134,8 +134,42 @@ def inference_to_file(
     output_file = output_path / filename
     concat_dataset.to_csv(index=False, path_or_buf=output_file)
 
+    raven_filename = f"raven_annotations-{datetime.datetime.now().strftime('%Y-%m-%d_%H-%M-%S')}-{model_name}-{dataset_name}-thresh03.csv"
+    raven_output_file = output_path / raven_filename
+    raven_df = inference_csv_to_raven(results_df, test_dataset, threshold=0.5)
+    raven_df.to_csv(index=False, path_or_buf=raven_output_file, sep="\t")
+
     return
 
+def inference_csv_to_raven(probsdataframe, test_dataset, threshold=0.5):
+    # transforms the probability dataframe to a raven format with class_1 predictions as the annotated bounding boxes
+    len_dataset = probsdataframe.shape[0] #number of segments in wav
+    seq_len = test_dataset.seq_length
+    all_begin_times = np.arange(0, len_dataset * seq_len, seq_len)
+    positives = []
+    for index, row in probsdataframe.iterrows():
+        if row['class1_prob'] > threshold:
+            positives.append(index)
+
+    begin_times = np.round(all_begin_times[positives],decimals=1)
+    end_times = np.round(begin_times+seq_len,decimals=1)
+    if end_times[-1] > round(len_dataset*seq_len,1):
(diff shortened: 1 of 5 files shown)


soundbay/inference.py
```diff
@@ -134,8 +134,45 @@ def inference_to_file(
     output_file = output_path / filename
     concat_dataset.to_csv(index=False, path_or_buf=output_file)
 
+    # save raven file
```
Z30G0D, line 137: Looks good,
We should add this to the hydra configuration .
(True/ False for saving it as raven file

soundbay/inference.py
```diff
@@ -134,8 +134,45 @@ def inference_to_file(
...
+    threshold = 0.5
```
Z30G0D, line 138: There is no use in this one since You define a default value in `inference_csv_to_raven` function.

Z30G0D (changes requested):
Looks great,
small comments

soundbay/inference.py
amitgalor18: good idea, will do

soundbay/inference.py
mosheman5: that's correct. Perhaps we should have inference args conf?

soundbay/inference.py
```diff
@@ -134,8 +134,45 @@ def inference_to_file(
...
+    thresholdtext = int(threshold*10)
+    raven_filename = f"raven_annotations-{datetime.datetime.now().strftime('%Y-%m-%d_%H-%M-%S')}-{model_name}-{dataset_name}-thresh0{thresholdtext}.csv"
+    raven_output_file = output_path / raven_filename
+    raven_df = inference_csv_to_raven(results_df, test_dataset, threshold=threshold)
+    raven_df.to_csv(index=False, path_or_buf=raven_output_file, sep="\t")
```
mosheman5, line 143: Does raven accept csv files as well? I thought it's only txt

soundbay/inference.py
```diff
@@ -134,8 +134,45 @@ def inference_to_file(
...
+
     return
 
+def inference_csv_to_raven(probsdataframe, test_dataset, threshold=0.5):
```
mosheman5, line 147: I think we should start include typing in our methods. 
Let's start with this pr :)

soundbay/inference.py
```diff
@@ -134,8 +134,45 @@ def inference_to_file(
...
+    # transforms the probability dataframe to a raven format with class_1 predictions as the annotated bounding boxes
+    len_dataset = probsdataframe.shape[0] #number of segments in wav
+    seq_len = test_dataset.seq_length
```
mosheman5, line 150: if the whole use of `test_dataset` is its attribute `seq_length`, I think `seq_length` should be the function argument and not `test_dataset`

soundbay/inference.py
```diff
@@ -134,8 +134,45 @@ def inference_to_file(
...
+    all_begin_times = np.arange(0, len_dataset * seq_len, seq_len)
+    positives = []
+    for index, row in probsdataframe.iterrows():
+        if row['class1_prob'] > threshold:
+            positives.append(index)
```
mosheman5, line 155: I think it would be easier to use 
```
if_positive = probsdataframe['class1_prob'] > threshold
begin_times = np.round(all_begin_times[if_positive],decimals=1)
```
instead of iterating the rows of the dataframe

soundbay/inference.py
```diff
@@ -134,8 +134,45 @@ def inference_to_file(
...
+
+    begin_times = np.round(all_begin_times[positives],decimals=1)
```
mosheman5, line 157: btw, why we need round or decimals=1? what's gonna happen we if want seq_length of 0.125?

soundbay/inference.py
```diff
@@ -134,8 +134,45 @@ def inference_to_file(
...
+    end_times = np.round(begin_times+seq_len,decimals=1)
+    if end_times[-1] > round(len_dataset*seq_len,1):
+        end_times[-1] = round(len_dataset*seq_len,1) #cut off last bbox if exceeding eof
+    low_freq = np.zeros_like(begin_times)
+    high_freq = np.ones_like(begin_times)*20000 #just tall enough bounding box
+    view = ['Spectrogram 1']*len(begin_times)
+    selection = np.arange(1,len(begin_times)+1)
+    channel = np.ones_like(begin_times)
+    channel = channel.astype(int)
```
mosheman5, line 166: small nit, but can be done in one line:
``` channel = np.ones_like(begin_times).astype(int) ```

soundbay/inference.py
amitgalor18: When I didn't add the rounding I got something messed up (e.g. 24.0000001 sec or something). Probably has something to do with the float variable

amitgalor18: When will we need such precise seq lengths? Is there significant data in the realm of 0.01 seconds?

soundbay/inference.py
amitgalor18: typing = type hints?
If so - I'm on it

soundbay/inference.py
amitgalor18: yup, it does. It was a marvelous discovery

soundbay/inference.py
amitgalor18: I added it to the conf file

soundbay/inference.py
amitgalor18: I tried it again and it seems that the non rounded numbers don't affect the visual result in raven. I'm not sure if it might affect evaluation calculations later, but if it will then we'll change it back.

PR #34, opened 2022-04-12
Add load_n_adapt_raven_annotation_table_to_dv_dataset_requirements

noa-weiss:


soundbay/utils/metadata_processing.py
```diff
@@ -8,6 +9,17 @@
...
 
+def load_n_adapt_raven_annotation_table_to_dv_dataset_requirements(annotation_path: str,
+                                                                   filename_suffix: str = ".Table.1.selections.txt"
+                                                                   ) -> pd.DataFrame:
+    # todo: decide whether to add annotation treatment
+    df_annotations = pd.read_csv(annotation_path, sep="\t")
+    df_annotations['filename'] = os.path.basename(annotation_path).replace(filename_suffix, '')
+    df_annotations = df_annotations.rename(columns={'Begin Time (s)': 'begin_time', 'End Time (s)': 'end_time'})
+    df_annotations['call_length'] = df_annotations['end_time'] - df_annotations['begin_time']
+    return df_annotations
```
mosheman5, line 20: Lgtm, do you think anything else is needed in here? @Z30G0D @amitgalor18 @noamzbr

mosheman5 (approved):

soundbay/utils/metadata_processing.py
Z30G0D: Maybe the function name can be slightly shorter? 
`adapt_raven_annotations_to_dataset`
And then elaborate in description.

Raven has a possibility to change the time formatting
image

I assume it's mostly in some kind of default that is similar to all of us.

But someone might want to mess with it.
So we need to verify the formats and the column's existence.
I suggest to use `assert` and [strptime](https://docs.python.org/3/library/datetime.html#strftime-strptime-behavior)

Is the TODO relevant to what I just wrote? 
If yes, sorry :)
Much needed feature, GJ!

Z30G0D (commented):
LGTM added comment on Moshe's comment

soundbay/utils/metadata_processing.py
noa-weiss: Re your second comment, I'm not sure I follow. What's the feature you're suggesting here?

mosheman5: Let's adapt this in the future if we'll find out researcher actually use it.
I can add it as tech debt task.
I don't think we need to prepare to all possible formats right now

Z30G0D: The idea was to validate current format (make sure that it is not changed from what we use).
But sounds good to me.

amitgalor18: Do you think maybe we should save additional information in the dataframe (e.g. frequency, class somehow) in case we want to use it in the future? Or is it a "when we reach the bridge, we cross it" type of deal?

amitgalor18 (approved):

soundbay/utils/metadata_processing.py
noa-weiss: @amitgalor18 I think the latter. It's easily to add once we need it.

PR #35, opened 2022-04-21
Feature/finetune per layer

rafaelsul123:
added freeze_layers method.
*currently, only implemented for ResNet1Channel based models.

added runs/main_finetune config, optim/jasco_resnet_50_finetune config
added config parameter "optim.num_unfrozen"

the method will freeze blocks of layers according to the number of num_unfrozen as defined within runs configuration file, in reverse order (when value = 0, no freezing
 occurs. when value = 1 -> freeze the last layer (FC). when value = 2 -> freeze FC + layer4 block, .... )


soundbay/train.py
```diff
@@ -44,6 +44,7 @@ def modeling(
     scheduler_args,
     model_args,
     logger,
+    num_unfrozen,
```
noamzbr, line 47: 1. Should have a default = None, in which case everything is unfrozen. 
2. Add this parameter to docstring (after logger)

soundbay/train.py
```diff
@@ -103,7 +104,12 @@ def modeling(
         logger=logger
     )
 
+    # Freeze layers if required (num_unfrozen!=0)
+    if num_unfrozen:
```
noamzbr, line 108: I think that the behaviour where `num_unfrozen == 0` means "everything can be trained" is confusing (I would expect it to mean "nothing can train", which makes no sense but is the natural way to interpert this parameter). I suggest that `num_unfrozen is None` would signify the "usual" case where nothing is frozen.

soundbay/train.py
```diff
@@ -176,6 +182,7 @@ def main(args):
         scheduler_args=args.optim.scheduler,
         model_args=args.model.model,
         logger=logger,
+        num_unfrozen=args.optim.num_unfrozen,
```
noamzbr, line 185: Please make sure - if running this with a regular model (no freezing), this arg will pass `None`?

soundbay/conf/optim/jasco_resnet_50_finetune.yaml
```diff
@@ -0,0 +1,12 @@
...
+  epochs: 100
+  optimizer:
+    _target_: torch.optim.SGD
+    lr: 0.001
+    momentum: 0.9
+    weight_decay: 0.0001
+  scheduler:
+    _target_: torch.optim.lr_scheduler.ExponentialLR
+    gamma: 0.95
+  num_unfrozen: 0 # 0 = No freeze, 1 = freeze FC,   2 = freeze FC+block4,   ...
```
noamzbr, line 12:
```suggestion
  num_unfrozen: 0 # 0 = No freeze, 1 = unfreeze FC,   2 = unfreeze FC+block4,   ...
```
And as suggested before, I think `num_unfrozen: 0` should mean "all are frezen"

noamzbr (commented):
Real nice catch with this bug! I have a few suggestions for this mechanism, but will be glad to hear more opinions.

mosheman5 (changes requested):
Looks good suggestion for the feature.
IMO we should not specify number of training layers, since it's not generic enough for other models. I also think that we don't need this extra degree of freedom.
Ideally, I would suggest instead of the `freeze_layers` method, each model to have
`prepare_model_for_finetune`. Which layers to finetune should be tested in an experiment and not for the user to decide. 
So instead of passing int argument of number of layers to freeze, I suggest moving to binary mode: either finetune the model or not.
Applying it on other models shouldn't be difficult in that case.

soundbay/train.py
rafaelsul123: I've changed the way this method and parameter behave according to Moshe's request (it now has True / False / None values)

about 2 - is printing a statement about being in finetune mode a "good enough" log? (shows up in "logs" in wandb)
image

soundbay/train.py
rafaelsul123: I'm afraid this implementation demands we add this argument to all optim configuration files.
I made sure it works as default 'False' even if the argument is only stated, with no attached value (in which case it passes "None" and is then turned into "False" but the framework doesn't work if the argument is none-existent in the optim file.

soundbay/models.py
```diff
@@ -29,6 +29,22 @@ def _choose_block_class(block):
         module_name = '.'.join(block.split('.')[:-1])
         return getattr(importlib.import_module(module_name), class_name)
 
+    def freeze_layers(self, freeze_layers_for_finetune=False):
+        """
+        Freeze all layers except the classifier and last layer block from training, as a condition for finetune
+        Input:
+            freeze_layers_for_finetune: True if objective is to finetune the model (default == False)
+        """
+        if freeze_layers_for_finetune:
```
mosheman5, line 38: looks great to me, the only thing I would change is removing the redundant `freeze_layers_for_finetune` flag
I mean, if the user calls model.freeze_layers() he probably want's to freeze layers, adding extra bool flag might be cumbersome at the happy flow or potential bug in the sad (and realistic..) flow.
So no args are needed inside `freeze_layers` imo

soundbay/conf/optim/jasco_resnet_50_finetune.yaml
```diff
@@ -0,0 +1,12 @@
...
+  epochs: 100
+  optimizer:
+    _target_: torch.optim.SGD
+    lr: 0.001
+    momentum: 0.9
+    weight_decay: 0.0001
+  scheduler:
+    _target_: torch.optim.lr_scheduler.ExponentialLR
+    gamma: 0.95
+  freeze_layers_for_finetune:  # Default is False. If True, freeze classifier + last layer/block
```
mosheman5, line 12: should it be
`freeze_layers_for_finetune: False`?
does it run without a value?

mosheman5: I also wonder if it's more a `optim` arg or `experiment`. I don't have a strong opinion, think either are fine

soundbay/models.py
rafaelsul123: Made the changes as requested

soundbay/models.py
```diff
@@ -29,6 +29,18 @@ def _choose_block_class(block):
...
+    def freeze_layers(self):
+        """
+        Freeze all layers except the classifier and last layer block from training, as a condition for finetune
+        """
+
+        for param in self.parameters():
+            param.requires_grad = False
+        for param in self.fc.parameters():
+            param.requires_grad = True
+        for param in self.layer4.parameters():
```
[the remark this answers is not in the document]
rafaelsul123: This should not happen, as the implementation above is specific to all ResNet1Channel based models (those should all have the above attributes). I think if a future implementation of a new model would require to remove a layer block (like layer4), or change the classifier (FC) / add additional blocks, etc - the child model should have the method overloaded and designed for its own architecture

the method for models not based on ResNet1Channel is not yet implemented (it would require testing before implementation, to hopefuly find optimal finetune setting)

mosheman5 (approved):

PR #37, opened 2022-04-24
use max volume of -6db for audio artifacts

mosheman5:


soundbay/utils/logging.py
```diff
@@ -152,8 +152,7 @@ def upload_artifacts(self, audio: torch.Tensor, label: torch.Tensor, raw_wav: to
         # Original wavs batch
 
         artifact_wav = torch.squeeze(raw_wav).detach().cpu().numpy()
-        artifact_wav = artifact_wav / np.max(np.abs(artifact_wav)) * 0.5 #avoid clipping
-        artifact_wav = artifact_wav * 10 ** (volume / (20 * np.log10(1/ np.std(artifact_wav)))) # add dB 
+        artifact_wav = artifact_wav / np.expand_dims(np.abs(artifact_wav).max(axis=1), 1) * 0.5  # gain -6dB
```
Z30G0D, line 155: Can `artifact_wav == 0`  in some edge case?

mosheman5: Yes, in that case we'll just log silent audio.
It won't break anything

mosheman5: Great catch, thanks!
Added a small number so we won't divide by zero

Z30G0D (approved):
lgtm

PR #38, opened 2022-05-01
Enable annotation files with channel, and additional fixes.

noamzbr:
The fixes:

1. Enable annotation files with a channel column. If not present, single channel will be assumed as previously.
2. Add the default of `resume="allow"` to wand.init args, enabling continuation of an existing run given providing the previous run id in `experiment.run_id`.
3. Fix an issue of start_time being selected before end_time at random. 
4. In `sf.read`, cast the path explicitly to string to cope with an issue that we have that looks like it's related to [this](https://github.com/librosa/librosa/issues/1019).

p.s - the bug that IMO neccessitates 4 is the following:
`RuntimeError: Error opening '../../FOLDER_NAME/FILE_NAME.wav': Error in WAV/W64/RF64 file. Malformed 'fmt ' chunk.`, though it's curious why his happned for this particular file only something like 100 epochs.


soundbay/conf/checkpoint/defaults.yaml
```diff
@@ -1,3 +1,4 @@
 # @package _global_
 checkpoint:
-  path: null
\ No newline at end of file
+  path: null
+  resume: 'allow'
```
mosheman5, line 4: Do we want this as a default flag?

soundbay/data.py
```diff
@@ -163,9 +167,17 @@ def _get_audio(self, path_to_file, begin_time, end_time, label):
                 start_time = random.randint(begin_time - margin_len_begin, end_time - margin_len_end)
             else:
                 start_time = random.randint(begin_time, end_time - int(self.seq_length * self.data_sample_rate))
+            if start_time < begin_time:
```
mosheman5, line 170: You actually turn off here the margin_len feature, not sure it's a good idea
what was the original bug that raised and issue?

soundbay/conf/checkpoint/defaults.yaml
noamzbr: As far as I understood this will resume the previous run as long as the previous run id is explicitly provided. Think this is the behavior we want as it's explicit enough. 
https://docs.wandb.ai/guides/track/advanced/resuming

soundbay/train.py
```diff
@@ -122,7 +122,7 @@ def main(args):
     _logger = wandb if not args.experiment.debug else Mock()
     experiment_name = get_experiment_name(args)
     _logger.init(project="finding_willy", name=experiment_name, group=args.experiment.group_name,
-                 id=args.experiment.run_id)
+                 id=args.experiment.run_id, resume=args.checkpoint.resume)
```
Z30G0D, line 125: Cool feature.
I know this in a different name in ClearML...

Z30G0D (approved):
Looks very helpful to me.
Anyone else?

PR #41, opened 2022-05-28
Feature/modify inference for multiclass

amitgalor18:
Refactored inference, extracting all the analysis functions (currently only metrics logging and creation of raven files) to a separate results_analysis.py.
Modified inference.py to accept multiclass.

Currently exporting the predicted labels by index (0,1,2, ...) and assuming the order in the test set is the same as the one the model was trained on.


soundbay/results_analysis.py
```diff
@@ -0,0 +1,108 @@
...
+    len_dataset = relevant_columns_df.shape[0]  # number of segments in wav
+    seq_len = seq_length
+    if_positive = probsdataframe[selected_class] > threshold  # check if the probability is above the threshold
+    if "begin_time" in probsdataframe.columns: #if the dataframe has metadata
+        all_begin_times = probsdataframe["begin_time"].values
+
+    else: #if the dataframe came from a file with no ground truth
+        all_begin_times = np.arange(0, len_dataset * seq_len, seq_len)
+
+    begin_times = all_begin_times[if_positive]
```
noamzbr, line 45: How do you know that probsdataframe and all_begin_times have the same shape? For e.g. if all_begin_times  was generated in the `else:` path

soundbay/results_analysis.py
```diff
@@ -0,0 +1,108 @@
...
+    # begin_times = np.round(all_begin_times[if_positive], decimals=1) #rounded to avoid float errors (e.g. 24.00000000000001)
+
+    end_times = np.round(begin_times+seq_len, decimals=3)
+    if end_times[-1] > round(len_dataset*seq_len,1):
+        end_times[-1] = round(len_dataset*seq_len,1) #cut off last bbox if exceeding eof
+    low_freq = np.zeros_like(begin_times)
+    high_freq = np.ones_like(begin_times)*20000 #just tall enough bounding box
```
noamzbr, line 52: upper freq shouldn't be hard coded.

soundbay/results_analysis.py
```diff
@@ -0,0 +1,108 @@
...
+
+    """
+    columns_array = results_df.columns[-int(num_classes):]
+    results_array = np.array(results_df[columns_array])
+    metrics_dict = Logger.get_metrics_dict(results_df["label"].values.tolist(),
+                                       np.argmax(results_array, axis=1).tolist(),
+                                       results_array)
+    print(metrics_dict)
+
+def inference_csv_to_raven(probsdataframe: pd.DataFrame, num_classes, seq_length: float, selected_class: str,threshold: float = 0.5, class_name: str = "call") -> pd.DataFrame:
```
noamzbr, line 32: Need to prettify this code a bit IMO

soundbay/results_analysis.py
amitgalor18: The `len_dataset` is taken from the length of the inference dataframe. In the `else` case it's just the length of the audio file (divided by seq_length) and in the `if` case it's a list of segments from the test set metadata

soundbay/results_analysis.py
amitgalor18: it was chosen for visibility in raven (so the bounding box is visible whether you're looking at a 0-1000 or a 0-100000Hz spectrogram), it has no real effect to my knowledge

soundbay/results_analysis.py
amitgalor18: done

soundbay/results_analysis.py
```diff
@@ -0,0 +1,117 @@
...
+    """
+    # configurations:
+
+    workdir = Path(os.getcwd())
+
+    output_dirpath = workdir.parent.absolute() / "outputs"
+    output_dirpath.mkdir(exist_ok=True)
+    inference_csv_name = "Inference_results-2022-05-03_17-01-10-best_margin0_5-220302_0054"
+    inference_results_path = output_dirpath / Path(inference_csv_name + ".csv")
+    num_classes = 2
```
Z30G0D, line 87: Hard coded

soundbay/results_analysis.py
```diff
@@ -0,0 +1,117 @@
...
+    return annotations_df
+
+def analysis_main() -> None:
+    """
+    The main function for running an analysis on a model inference results for required Dataset.
+
+    """
+    # configurations:
+
+    workdir = Path(os.getcwd())
```
Z30G0D, line 81: We should some minimal arguments (must include inference_file+output path at least)

Z30G0D (commented):
LGTM,
lots of code work there.
GJ Amit and Yogev

soundbay/results_analysis.py
amitgalor18: good idea, I'll add an assert statement to check that

soundbay/results_analysis.py
```diff
@@ -0,0 +1,130 @@
...
+
+    """
+    # configurations:
+    args = make_parser().parse_args()
+    workdir = Path(os.getcwd())
+    assert args.filename, "filename argument is empty, you can't run analysis on nothing"
+    output_dirpath = Path(args.filedir) #workdir.parent.absolute() / "outputs"
+    output_dirpath.mkdir(exist_ok=True)
+    inference_csv_name = args.filename
+    inference_results_path = output_dirpath / Path(inference_csv_name + ".csv")
```
Z30G0D, line 99: Doesn't this do ".csv.csv"?

soundbay/results_analysis.py
```diff
@@ -0,0 +1,130 @@
...
+
+def analysis_main() -> None:
+    """
+    The main function for running an analysis on a model inference results for required Dataset.
+
+    """
+    # configurations:
+    args = make_parser().parse_args()
+    workdir = Path(os.getcwd())
+    assert args.filename, "filename argument is empty, you can't run analysis on nothing"
```
Z30G0D, line 95: `workdir` is not in use.

soundbay/results_analysis.py
```diff
@@ -0,0 +1,130 @@
...
+    results_df = pd.read_csv(inference_results_path)
+    name_col = args.selected_class  # selected class for raven results
+
+    # go through columns and find the one containing the selected class
+    column_list = results_df.columns
+    selected_class_column = [s for s in column_list if name_col in s][0]
+    seq_length_default = 0.2  # relevant only if analyzing an inference single file (not a dataset)
+    seq_length = results_df["call_length"].unique()[0] if "call_length" in results_df.columns else seq_length_default  # extract call length from inference results
+
+    save_raven = True
```
Z30G0D, line 111: inside args, default should be true

soundbay/results_analysis.py
```diff
@@ -0,0 +1,130 @@
+import pandas as pd
+import torch
```
Z30G0D, line 2: Remove not needed imports

soundbay/results_analysis.py
```diff
@@ -0,0 +1,130 @@
...
+                      threshold: the threshold to be used for the selection.
+                      class_name: the name of the class for which the raven csv file is generated.
+
+        Returns: a pandas dataframe containing the raven csv file.
+    """
+
+    relevant_columns = probs_df.columns[-int(num_classes):]
+    relevant_columns_df = probs_df[relevant_columns]
+    len_dataset = relevant_columns_df.shape[0]  # number of segments in wav
+    seq_len = seq_length
```
Z30G0D, line 60: I guess it's needed :)

Z30G0D (changes requested):
Wonderful PR.
Well documented!

soundbay/results_analysis.py
amitgalor18: I checked, it doesn't

PR #42, opened 2022-06-05
Dataloader_toward_multiclass

Z30G0D:
Multiclass dataset
Not many changes are needed.

Make sure to use these during launch (and the mock dataset):
```
"data=defaults_multiclass",
"model.model.num_classes=4"
```


datasets/mozambique_2018/test_multiclass.csv
```diff
@@ -0,0 +1,51 @@
+Unnamed: 0,begin_time,end_time,filename,call_length,label
```
noamzbr, line 1: This file should better go in tests/assets/annotations

soundbay/data.py
```diff
@@ -80,18 +81,16 @@ def _preprocess_metadata(self, len_buffer, equalize=False, slice_flag=False):
...
-                return df_object
-            label = 0 if diff > 0 else 1
-            multiplier = int(np.ceil(abs(diff) / min(len_neg, len_pos)))
-            additive = pd.concat(deepcopy([self.metadata[self.metadata['label'] == label]]) * multiplier)
-            additive = additive[:abs(diff)]
-            df_object = pd.concat([df_object, additive])
-            assert len(df_object[df_object['label'] == 1]) == len(df_object[df_object['label'] == 0])
-            return df_object
+            # TODO: How should we consider background?
+            oversample = RandomOverSampler(sampling_strategy='auto')
```
noamzbr, line 85: Nice find. In the future perhaps we'll want an option to pass the input to this parameter from conf (to enable un-balanced sampling with a dict for e.g.)

soundbay/data.py
```diff
@@ -232,6 +231,46 @@ def __len__(self):
...
+    '''
+    This  class inherits all the traits from ClassifierDataset and handles cases with multiple classes
+    '''
+
+    
+    def _get_audio(self, path_to_file, begin_time, end_time, label):
+        """
+        _get_audio gets a path_to_file from _grab_fields method and also begin_time and end_time
+        and returns the audio segment in a torch.tensor. it differs from ClassifierDataset _get_audio method
+        by turning the first conditional into "not background" instead of "if call"
```
noamzbr, line 246: Why isn't this function suitable for the binary case?

soundbay/data.py
```diff
@@ -232,6 +231,46 @@ def __len__(self):
...
+        end_time - int
+
+        output:
+        audio - pytorch tensor (1-D array)
+        """
+        
+
+        if (self.mode == "train") and (label != 0):
+            if self.margin_ratio != 0:  # ranges from 0 to 1 - indicates the relative part from seq_len to exceed call_length
+                margin_len_begin = int((self.seq_length * self.data_sample_rate) * self.margin_ratio)
```
noamzbr, line 260: How do we avoid getting other non background signals inside our margins?

soundbay/trainers.py
```diff
@@ -99,6 +100,9 @@ def train_epoch(self, epoch):
 
             # estimate and calc losses
             estimated_label = self.model(audio)
+            if estimated_label.shape[1] > 2: #Softmax in case of multiclass
```
noamzbr, line 103: From reading https://pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html it seems like it should want to get un-normalized inputs (no softmax) for any number of classes.
In case I'm wrong here and the softmax is needed - need to add it also in the val loop.

Z30G0D: Right!
removed.

soundbay/data.py
Z30G0D: `(label != 0)` is a different condition then the original one `(label == 1)`
Now you mention it,
I can do it more elegant to avoid code replication

amitgalor18 (approved):
LGTM. 
I concur with Noam's comment about taking margin when there could be a different call nearby, we should decide what to do in this case and until then refrain from using 'margin_ratio>0' for multiclass datasets.

soundbay/data.py
mosheman5: yes, at least for the "only calls" setup where we don't have a specified background call

soundbay/data.py
```diff
@@ -232,6 +217,46 @@ def __len__(self):
         return self.metadata.shape[0]
 
 
+
+
+class MultiClassifierDataset(ClassifierDataset):
```
mosheman5, line 222: I think the naming is quite confusing. Why call is `MultiClassifierDataset` and `ClassifierDataset`, if the `ClassifierDataset` doesn't support more than two classes?
Or does it?

mosheman5: I would suggest a way for solving this issue, if I may:
having a base class dataset that is called `BaseDataset` with all the methods implemented except of `_get_audio` (that will raise NotImplementedError exception)
We'll have two classes to use:
`ClassifierDataset` - a dataset to be used with a background label, make sure it's label 0. The current implementation is actually the same, only we have the bug @noamzbr stated with taking margin from other call. We'll need to have some logic to make sure we only take from bg. As far as I'm concerned, we can keep the current one with a technical debt and implement it in the future, so we could merge the current PR asap.
`NoBackgroundClassifierDataset` - The same but this time there is no margin at all, just sampling from the calls. This dataset is gonna be used for training on Liz's data

Z30G0D:
We need to extend our tests a bit.
Found bug [here](https://github.com/deep-voice/soundbay/pull/42/commits/7a353e396a77e1783303cd62dd17a3015f376faf)
that sneaked into `master` branch and actually the default config wouldn't work.

And also another potential one [here](https://github.com/deep-voice/soundbay/pull/42/commits/d4fd444c38caaf3ebf184ed1a23daecd9ce4f83d)

It should be scheduled to another time though.

rafaelsul123:
> We need to extend our tests a bit. Found bug [here](https://github.com/deep-voice/soundbay/pull/42/commits/7a353e396a77e1783303cd62dd17a3015f376faf) that sneaked into `master` branch and actually the default config wouldn't work.
> 
> And also another potential one [here](https://github.com/deep-voice/soundbay/pull/42/commits/d4fd444c38caaf3ebf184ed1a23daecd9ce4f83d)
> 
> It should be scheduled to another time though.

I was wondering what caused this bug in my local setup that used to work, good catch! I would've wasted a while trying things out if it wasnt for your comment and timing. thanks!

Z30G0D:
> > We need to extend our tests a bit. Found bug [here](https://github.com/deep-voice/soundbay/pull/42/commits/7a353e396a77e1783303cd62dd17a3015f376faf) that sneaked into `master` branch and actually the default config wouldn't work.
> > And also another potential one [here](https://github.com/deep-voice/soundbay/pull/42/commits/d4fd444c38caaf3ebf184ed1a23daecd9ce4f83d)
> > It should be scheduled to another time though.
> 
> I was wondering what caused this bug in my local setup that used to work, good catch! I would've wasted a while trying things out if it wasnt for your comment and timing. thanks!

I am glad this proved to be useful so soon  :)

PR #46, opened 2022-07-19
Can now load sections that are larger than calls and label them as calls

noamzbr:
Also, returned data equalization. why did we remove it?

### DIFF
--- a/soundbay/data.py
+++ b/soundbay/data.py
@@ -23,8 +23,7 @@ class BaseDataset(Dataset):
     '''
     def __init__(self, data_path, metadata_path, augmentations, augmentations_p, preprocessors,
                  seq_length=1, data_sample_rate=44100, sample_rate=44100, mode="train",
-                 equalize_data=False, slice_flag=False, margin_ratio=0,
-                 split_metadata_by_label=False):
+                 slice_flag=False, margin_ratio=0, split_metadata_by_label=False):
         """
         __init__ method initiates ClassifierDataset instance:
         Input:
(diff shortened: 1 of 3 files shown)


Z30G0D:
> Also, returned data equalization. why did we remove it?

The equalizing distribution (or resampling actually) is now part of `train.py`.
The `DataLoader` is responsible for this operation and not a method of `DataSet`.

soundbay/data.py
```diff
@@ -220,25 +244,42 @@ def _get_audio(self, path_to_file, begin_time, end_time, label, channel=None):
...
-            else:
-                start_time = random.randint(begin_time, end_time - int(self.seq_length * self.data_sample_rate))
-            if start_time < 0:
-                start_time = 0
+        seg_length = end_time - begin_time
+        requested_seq_length = int(self.seq_length * self.data_sample_rate)
+        last_start_time = sf.info(path_to_file).frames - requested_seq_length
+        # Do all this stuff only to calls in training set, because otherwise _slice_sequence has already been done
+        if self.mode == "train":
+            if seg_length > requested_seq_length:
```
mosheman5: The flow looks legit, good bug catch with the start time of label==0!

tests/test_data.py
```diff
@@ -14,7 +14,7 @@ def test_dataloader() -> None:
                                         augmentations=cfg.data.train_dataset.augmentations,
                                         augmentations_p=cfg.data.train_dataset.augmentations_p,
                                         preprocessors=cfg.data.train_dataset.preprocessors)
-        assert data_loader.metadata.shape[1] == 5  # make sure metadata has 5 columns
+        assert data_loader.metadata.shape[1] in (5, 6)  # make sure metadata has 5/6 columns (account for channel)
```
mosheman5, line 17: like for maintaining the tests

mosheman5 (approved):
Ahla feature

PR #47, opened 2022-07-19
Experiment/active learning validation

noa-weiss:


datasets/orcasound/adapt_active_learning_annotations.ipynb
```diff
@@ -0,0 +1,1909 @@
+{
```
amitgalor18, line 1: Is the whole notebook just one large dict\some data object?

amitgalor18 (approved):
Since it seems that the whole pr is just additional utility files for active learning I see no reason not to approve it

noamzbr:
I lowkey believe that stuff that is in the repo has to be related to training and serving models, so that active learning related code does fit but the annotation produced by it belong more in s3/clickup/someotherstorage, but this is not a strongly held belief (:

noa-weiss:
@noamzbr , I'm willing to go with the general opinion here. @mosheman5 , WDYT?

noa-weiss:
> @noamzbr , I'm willing to go with the general opinion here. @mosheman5 , WDYT?

Paging Dr. @mosheman5

datasets/orcasound/adapt_active_learning_annotations.ipynb
```diff
@@ -0,0 +1,1909 @@
...
+ "cells": [
+  {
+   "cell_type": "markdown",
```
mosheman5, line 4: Can you please clear the outputs of the notebook before committing it to git? :) 
https://stackoverflow.com/questions/28908319/how-to-clear-jupyter-notebooks-output-in-all-cells-from-the-linux-terminal
you can also do it directly inside the notebook using the GUI

noa-weiss:
Could you explain why?

mosheman5:
> 

Generally speaking outputs are static objects, and they are leaves of the process.
Ie, they are not used in any other process down the line, and we need no diff logging for them (unlike metadata objects/csv that are needed for training a model).
The main disadvantage of saving output data is unnecessary enlarging the repo.
Ideally the code should be slim, and if additional data is generated/needed we put references to its location in the code instead of the data itself.

PR #48, opened 2022-07-23
Feature/data min freq

rafaelsul123:
implemented minimal frequency cutoff for spectrogram - takes place within the preprocessing, as part of _preprocessors.yaml config (after spectrogram, before amplitude_2_db)
the argument is called min_freq_spectrogram and is input through data config.
updated all data training configs with min_freq_spectrogram=0 - change as required.

### DIFF
--- a/soundbay/conf/preprocessors/_preprocessors.yaml
+++ b/soundbay/conf/preprocessors/_preprocessors.yaml
@@ -4,6 +4,11 @@ _preprocessors:
     _target_: torchaudio.transforms.Spectrogram
     n_fft: ${data.n_fft}
     hop_length: ${data.hop_length}
+  min_freq:
+    _target_: soundbay.data.MinFreqSpectrogram
+    min_freq_spectrogram: ${data.min_freq_spectrogram}
+    sample_rate: ${data.sample_rate}
+    n_fft: ${data.n_fft}
   amplitude_2_db:
     _target_: torchaudio.transforms.AmplitudeToDB
   peak_norm:
--- a/soundbay/conf/preprocessors/_preprocessors.yaml
+++ b/soundbay/conf/preprocessors/_preprocessors.yaml
@@ -4,9 +4,9 @@ _preprocessors:
     _target_: torchaudio.transforms.Spectrogram
     n_fft: ${data.n_fft}
     hop_length: ${data.hop_length}
-  min_freq:
-    _target_: soundbay.data.MinFreqSpectrogram
-    min_freq_spectrogram: ${data.min_freq_spectrogram}
+  min_freq_filtering:
+    _target_: soundbay.data.MinFreqFiltering
+    min_freq_filtering: ${data.min_freq_filtering}
     sample_rate: ${data.sample_rate}
   amplitude_2_db:
     _target_: torchaudio.transforms.AmplitudeToDB
--- a/soundbay/conf/preprocessors/_preprocessors.yaml
+++ b/soundbay/conf/preprocessors/_preprocessors.yaml
@@ -8,7 +8,6 @@ _preprocessors:
     _target_: soundbay.data.MinFreqSpectrogram
     min_freq_spectrogram: ${data.min_freq_spectrogram}
     sample_rate: ${data.sample_rate}
-    n_fft: ${data.n_fft}
   amplitude_2_db:
     _target_: torchaudio.transforms.AmplitudeToDB
   peak_norm:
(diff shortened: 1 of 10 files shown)


soundbay/data.py
```diff
@@ -288,6 +290,25 @@ def __call__(self, sample):
...
 
 
+class MinFreqSpectrogram:
+    """Cut the spectrogram frequency axis to make it start from min_freq"""
+    def __init__(self, min_freq_spectrogram, sample_rate):
+        self.min_freq_spectrogram = min_freq_spectrogram
+        self.sample_rate = sample_rate
+
+    def edit_spectrogram_axis(self, sample):
+        max_freq_in_spectrogram = self.sample_rate / 2
```
yogevmatalon, line 300: maybe add a comment for using the max_freq parameter if it will be implemented in the future

yogevmatalon (changes requested):
As discussed, need to check how this change affects the inference. instances with lower freq will be removed, tagged as 0, or regular prediction by the model?

noamzbr:
> As discussed, need to check how this change affects the inference. instances with lower freq will be removed, tagged as 0, or regular prediction by the model?

I understand that this needs to be present also in the inference pipeline, but I'm not sure I followed your comment about how it will affect instances. I assumed that this stage is idependant of tagging, and that it's the user's (our) responsaiblity to make sure we're not cutting our our calls.

soundbay/conf/preprocessors/_preprocessors.yaml
```diff
@@ -4,6 +4,10 @@ _preprocessors:
     _target_: torchaudio.transforms.Spectrogram
     n_fft: ${data.n_fft}
     hop_length: ${data.hop_length}
+  min_freq:
```
Z30G0D, line 7: I think we should make it more informative.
`min_freq_filtering`?

soundbay/data.py
```diff
@@ -288,6 +290,25 @@ def __call__(self, sample):
         return (sample - sample.min()) / (sample.max() - sample.min())
 
 
+class MinFreqSpectrogram:
+    """Cut the spectrogram frequency axis to make it start from min_freq"""
+    def __init__(self, min_freq_spectrogram, sample_rate):
```
Z30G0D, line 295: need to add type check in function definition.
`sample_rate` is integer , `min_freq_spectrogram` is `Tensor`(I Think?)

Z30G0D (changes requested):
Very handy.
Great job!
+1 for @yogevmatalon 's and Noam's comment.

soundbay/conf/preprocessors/_preprocessors.yaml
rafaelsul123: sounds right, changed the name

soundbay/data.py
rafaelsul123: as of current implementation, inference is not affected in any way. I can implement the filtering for inference as well, but I think that deserves a seperate implementation that can be controlled through the relevant conf files (in parallel to the training conf files and implementation)

given the current implementation (and assuming no major changes), the inference will act just like normal, using spectrograms from freq=0 to sample_rate/2, and the output will depend solely on the trained model's ability to detect the signal, wherever it is within the spectrum

*about the comment, i've added a note within data.py, inside class implementation, and also added checks to make sure that  0 < min_freq_filtering < sample_rate/2
if/when we implement a max_freq filter similar to the min_freq one, id suggest we implement a check that takes place in both of the proccesses to make sure there are no contradictions

rafaelsul123:
hey guys, would love a second check, I tried to answer all the requested changes and would love to PR the feature :)

noamzbr (approved):
Approved but would like to know/add two things:
1. Just making sure - did you visually inspect the result to see that the `:-min_value` thing is the right way to go? I have no inturition why the spectorgram should start with 0 being the max freq...
2. Please also add `min_freq_filtering` to `_preprocesors_unit`

rafaelsul123:
> Approved but would like to know/add two things:
> 
> 1. Just making sure - did you visually inspect the result to see that the `:-min_value` thing is the right way to go? I have no intuition why the spectorgram should start with 0 being the max freq...
> 2. Please also add `min_freq_filtering` to `_preprocesors_unit`

1. Checked and works as intended (saved jpg's to make sure). I agree it doesnt make sense considering the math behind the spectrogram, but is does work as intended considering the Tensor data structure (the 0,0 point is top left corner, same as np arrays)
2. Added. btw, the n_fft and hop_length there are hard coded, is this intended?

Z30G0D (approved):

yogevmatalon (approved):

PR #50, opened 2022-08-05
Fix matplotlib memory leak in artifact upload

noamzbr:
put figure creation and deletion inside a loop to try and prevent memory leak


noamzbr:
Can confirm that it solved memory leak.
See memory usage in [pre-fix run](https://wandb.ai/deep_voice/finding_willy/runs/3t2ruxpv/system?workspace=user-deep_voice) vs [post-fix run](https://wandb.ai/deep_voice/finding_willy/runs/37pcp3mk/system?workspace=user-deep_voice)

mosheman5 (approved):

PR #51, opened 2022-08-16
Infer with channel

noamzbr:
Modify inference code such the inference is done per each channel in the audio file


soundbay/conf/preprocessors/_preprocessors_unit.yaml
```diff
@@ -2,8 +2,8 @@
 _preprocessors:
   spectrogram:
     _target_: torchaudio.transforms.Spectrogram
-    n_fft: 1024
-    hop_length: 256
+    n_fft: ${data.n_fft}
+    hop_length: ${data.hop_length}
```
amitgalor18, line 6: did you change the conf files in the data folder to include these parameters? So it won't break experiments who use them

soundbay/inference.py
```diff
@@ -109,36 +112,60 @@ def inference_to_file(
...
     dataset_name = Path(test_dataset.metadata_path).stem
     filename = f"Inference_results-{datetime.datetime.now().strftime('%Y-%m-%d_%H-%M-%S')}-{model_name}-{dataset_name}.csv"
     output_file = output_path / filename
-    concat_dataset.to_csv(index=False, path_or_buf=output_file)
-
-    # save raven file
+    pd.concat(all_channel_list, axis=0).to_csv(index=False, path_or_buf=output_file)
+
+    # Save raven file
+    if save_raven:
```
amitgalor18, line 162: didn't we move this section to "results_analysis.py"? are you sure you took the most updated version from master? I might be wrong though.

soundbay/results_analysis.py
```diff
@@ -67,11 +65,11 @@ def inference_csv_to_raven(probs_df: pd.DataFrame, num_classes, seq_len: float,
 
     # create columns for raven format
     low_freq = np.zeros_like(begin_times)
-    high_freq = np.ones_like(begin_times)*bb_height
+    high_freq = np.ones_like(begin_times)*max_freq
```
amitgalor18, line 68: good fix, I didn't like my hacky way to handle the boxes.

amitgalor18 (approved):
Good pr. I left some comments, check if they're relevant because I might be confused

soundbay/conf/preprocessors/_preprocessors_unit.yaml
noamzbr: The normal preprocessor.yaml file already has that, just updating preprocessing_unit.
Also datafiles do have these parameters.

soundbay/inference.py
noamzbr: in results_analysis inference_csv_to_raven is just saved as is. 
He're I'm merging the results from the different channels and re-setting seelction, so I have more code.
Anyaway using inference_csv_to_raven in both cases

soundbay/inference.py
```diff
@@ -109,36 +112,60 @@ def inference_to_file(
...
+                                                   results_df.values)
+            print(metrics_dict)
+        else:
+            concat_dataset = results_df
+            print("Notice: The dataset has no ground truth labels")
+
+        # create raven file
+        if save_raven:
+            all_channel_raven_list.append(
+                inference_csv_to_raven(results_df, predict_prob.shape[1], dataset_args.seq_length, 1, threshold, 'call',
```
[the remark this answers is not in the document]
noamzbr: fixed

noamzbr (commented):
A bit of an ugly fix, but the mechanism for supporting single file datasets and regular datasets are not the same.

soundbay/inference.py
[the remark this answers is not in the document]
sngImagry: Hey there.
I think we still have a problem - really small fix is needed; single tab is missing to make the hierarchy right in `inference_file.yaml` OR using `data.seq_length` instead of `data.test_dataset.seq_length` in the code snippet that is linked to the comment.
I know this PR has already been merged but I am writing here for documentation purposes

PR #52, opened 2022-09-03
Transition to using audiomentations rather than our own custom augmentations

noamzbr:
audiomentations contains all our existing augmentations (with a sligher better implementation) togheter with multiple additional augmentations (can be found [here](https://iver56.github.io/audiomentations/waveform_transforms/)).

For now only replaced the existing augs with their augs and added PitchShift, but can add way more (e.g. convolute with some impulse response function, add room acoustics and more).

I suggest that future custom augmentations (such as adding real background noise) will be subclased from their augmentation base class.


soundbay/data.py
```diff
@@ -151,7 +151,9 @@ def _set_augmentations(self, augmentations_dict, augmentations_p):
             augmentations_list = [instantiate(args) for args in augmentations_dict.values()]
         else:
             augmentations_list = []
-        augmenter = ChainedAugmentations(augmentations_list, augmentations_p) if self.mode == 'train' else torch.nn.Identity()
+        augmenter = lambda x: torch.tensor(
+            Compose(augmentations_list, p=augmentations_p, shuffle=True)(x, self.sample_rate)
```
Z30G0D, line 155: Nice, does the `shuffle` feature here should be something we should attend compared to previous version?

Z30G0D (commented):
Nothing much to add.
Did you try running a training using these?

soundbay/conf/augmentations/_augmentations.yaml
```diff
@@ -1,20 +1,24 @@
 # @package _global_
 _augmentations:
   random_noise:
-    _target_: soundbay.data_augmentation.AddGaussianNoise
+    _target_: audiomentations.AddGaussianSNR
```
rafaelsul123, line 4: Did you run an experiment to see how these audiomentations work compared to our "self built" methods?

rafaelsul123 (commented):
looks good

soundbay/data.py
noamzbr: I think there is no harm in adding this, but we don't have to. Can start without it

soundbay/conf/augmentations/_augmentations.yaml
noamzbr: Not yet, my next task is compiling a set of augs I belive can do better than our current augs. This is just an infrastructure change.

Z30G0D (approved):

PR #54, opened 2022-10-29
Two fixes

noamzbr:
1. Fix inference on files with no predictions
2. Reduce amount of audio/spectrograms logged


Z30G0D:
LGTM.
Nothing to add really.
Maybe we can change the epoch interval of 5 to behave as a variable , nothing more than this.

Z30G0D (approved):
LGTM.
Good catch!
Nothing to add really.
Maybe we can change the epoch interval of 5 to behave as a variable , nothing more than this.

PR #59, opened 2023-02-11
Fix tests

Z30G0D:
Fix tests and some warnings (as we failed to check them in the previous PR).

There are many warnings, maybe we should consider attending (or not :) )

Edit:
Adding elaborate description on how to run the configuration:
`@hydra.main(config_name="/runs/main", config_path="conf", version_base='1.2')`
is the wrapping function for our main training function (we have the same wrapper on the `inference.py`).

`config_path` argument contains the "root" of all the configs in order to access all of our conf `yaml` files.
`config_name` contains the [defaults_list](https://hydra.cc/docs/advanced/defaults_list/#:~:text=The%20Defaults%20List%20is%20a,a%20part%20of%20output%20config.) file, this is actually an "external" file that build the final configuration file from all the different `yaml`.

Note that I changed `main.yaml` keys to be `../**`, this is a result of the configuration files locationa that is outiside the `runs` folder and inside their respective folders **relative** to `main.yaml`.

In previous Hydra versions, it was slightly different (`requirements.txt` is updated with `hydra-core=1.2.0` that you need to update).

If you want to create a new `yaml` file (e.g. in `conf/data` folder) make sure that the top part of the `yaml` still contains the word data, since this is the `key` you will access when overriding through the command line (`train.py data=new_conf_data_file_name`).

The final configuration is encapsulated in the `validate_args` variable in `train.py`.

Later on,  there is an in-place operation (`resolve` by OmegConf) that resolves all the values ( turning all the $ in the config files into their respective values for example).

`Config(**validate_args)`  is the class validation operation by `pydantic`.
The class takes the resolved `dict` variable and makes sure that the fields are correct from configuration guidelines POV (models are allowed, certain configuration constraints etc..).

The `conf_validator.py` (file that contains `Config` class) file is mostly based on the following article.
https://suneeta-mall.github.io/2022/03/15/hydra-pydantic-config-management-for-training-application.html

I think it's pretty straightforward, just need to go through the [pydnatic](https://docs.pydantic.dev/) documentation to understand some of the stuff.


sngImagry (approved):
What is the expected base directory now? maybe you can add in the release notes what should be the directories structure from now on, given this PR and the one before it?
Something like this:

```
project
│   train.py
│
└───conf
│   │
│   └───runs
│      │   main.yml  # returns to be under `runs`
│      │   inference.yml
│   │
│   └───model
│      │   file11.yml
│       │   file12.yml
│       │   ...
│   
...
```
(I do want to pull the changes into my branch but structure changes are usually harder (for me) to merge, especially with relative paths. It never works out of the box so documentation can help out here...).

shaing10 (approved):

Z30G0D:
@shaing10 I changed the conf structure library.
This should be working as you intended, no?

shaing10 (approved):
Thanks!

Z30G0D:
@amitgalor18 I edited the first comment in the PR to explain the stuff you wanted (I think),
Let's talk in private if more clarifications are needed.

mosheman5 (commented):
Looks good, includes important changes
plz merge asap since this pr solves few bugs in the repo

PR #62, opened 2023-03-04
Tomer/fixing inference multiple files

Z30G0D:
I am not really sure about the `defaults_inference.yml` file.


amitgalor18 (approved):
looks good overall, but as I understand it all of the files in "runs" (main, main_inference etc.) need to be outside of the subfolder hierarchy, right? If so then we can extract everything from runs to the parent directory and delete the folder altogether

mosheman5 (approved):

PR #65, opened 2023-04-09
[dataloader] added a train_as_val data for evaluation

sngImagry:
data from the train set which is processed as validation set


soundbay/train.py
```diff
@@ -95,6 +104,8 @@ def modeling(
     model = models_dict[model_args.pop('_target_')](**model_args)
 
 
+    print('*** model has been loaded successfully ***')
+    print(f'number of trainable params: {sum([p.numel() for p in model.parameters() if p.requires_grad]):,}')
```
mosheman5, line 108: like!

soundbay/trainers.py
```diff
@@ -70,7 +72,8 @@ def train(self):
             self.logger.reset_losses()
             self.train_epoch(epoch)
             self.epochs_trained += 1
-            self.eval_epoch(epoch)
+            self.eval_epoch(epoch, 'val')
+            self.eval_epoch(epoch, 'train_as_val')
```
mosheman5, line 76: perhaps we could add flag for `eval_train_as_val_interval` so we won't have to go over the whole train set every epoch, because it might be much larger than the validation set. I think a default of once per 10 epoch could suffice, wdyt?

mosheman5 (changes requested):
Looks great! Great feature for improving our model and training analysis.
I suggest adding a parameters for testing the train only once in x epochs so we won't lose much time training the models.

mosheman5 (approved):

PR #66, opened 2023-05-29
Feature/add googlenet

amitgalor18:
Adding googlenet model to models, with a transform to the spectrograms to make them 3 channeled (RGB) in order to use a pretrained googlenet, based on the JASA paper. Creating conf files to enable experiments with the googlenet model


soundbay/conf/experiment/inference.yaml
```diff
@@ -1,7 +1,7 @@
 # @package _global_
 experiment:
   checkpoint:
-    path: ???
+    path: '../checkpoints/1dyeumy3/best.pth'
```
Z30G0D, line 4: This should not be commited

soundbay/conf/main.yaml
```diff
@@ -0,0 +1,8 @@
+# @package _global_
```
Z30G0D, line 1: Why is this file exist?
Should there be so many new `yaml` files?

soundbay/utils/conf_validator.py
```diff
@@ -1,7 +1,7 @@
 from typing import Optional
 from pydantic import BaseModel, validator
 from conf_dict import datasets_dict, criterion_dict, models_dict 
-
+import pdb
```
Z30G0D, line 4: Need to remove

soundbay/utils/logging.py
```diff
@@ -173,20 +173,26 @@ def upload_artifacts(self, audio: torch.Tensor, label: torch.Tensor, raw_wav: to
...
-        list_of_specs_objects = [wandb.Image(data_or_path=spec, caption=f'label_{lab}_{ind}_train') for spec, ind, lab in zip(specs,idx, label)]
-        log_wavs = {f'First batch {flag} original wavs': list_of_wavs_objects}
-        log_specs = {f'First batch {flag} augmented spectrogram\'s': list_of_specs_objects}
-
-        # Upload to W&B
-        wandb.log(log_wavs, commit=False)
-        wandb.log(log_specs, commit=False)
+        #avoid spectrogram upload if using 3 channel spectrograms (in googlenet), upload only wavs
+        if audio.size()[1] == 3:
+            log_wavs = {f'First batch {flag} original wavs': list_of_wavs_objects}
```
Z30G0D, line 178: Need to notify user somehow about this, he will be surprise.
Maybe use a warning or `info` statement to notify about this.
Should appear only once or twice during the run though.

soundbay/utils/signal_processing.py
```diff
@@ -66,3 +66,30 @@ def __call__(self,  sample, gain=0.6, bias=0.1, power=0.2, time_constant=0.4, ep
 
 
         return torch.from_numpy(pcen_librosa).float()
+
+class DuplicateChannels(object):
```
Z30G0D, line 70: Looks very good!

Z30G0D (changes requested):
Great PR!
Important, your test seems to be failing.
Try to consult this post:
https://stackoverflow.com/questions/70998767/no-module-named-torchvision-models-utils

soundbay/conf/experiment/inference.yaml
amitgalor18: Agree, will be removed

soundbay/conf/main.yaml
amitgalor18: I think that different types of experiments (e.g. with many different parameters such as model, preprocessing, optimizer, and not just one different parameter) are easer to handle when they have their own main. If it's only one parameter like a dataset path it's simple enough to add an override to the train command line.  The new "optim", "preprocessors" and "model" yamls are of course required.

But on a related note, I may have misunderstood the changes you made in the latest PR to the hierarchy, I thought that every "runs" conf file should now be in the parent folder in order to reach all the relevant yamls, so I moved main, main_inference and the parallel files for the googlenet manatees experiment up to the conf folder.

soundbay/utils/logging.py
amitgalor18: good idea, I'll check how to do that

PR #67, opened 2023-05-29
fix to data.py to avoid bug in some datasets

amitgalor18:
added necessary condition to multichannel in data.py to allow datasets with 1 channel but existing channel column in metadata


soundbay/data.py
```diff
@@ -263,7 +263,7 @@ def _get_audio(self, path_to_file, begin_time, end_time, label, channel=None):
             start_time = begin_time
         data, _ = sf.read(str(path_to_file), start=start_time,
                           stop=start_time + requested_seq_length)
-        if channel is not None:
+        if channel is not None and channel>1:
```
mosheman5, line 266: don't we want to check the shape of data instead of the value of channel?
If I have a multichannel input I still want to be able to read data from channel 1.

Z30G0D: Sometimes you have another channel that is empty.

PR #72, opened 2023-06-17
Feature/dataset creator final

rafaelsul123:
The dataset_creator.py file should be a relatively simple way to build new datasets out of annotation files, rather than the tidious process of running through notebooks, creating and editing many cells depending on the specific dataset.
The script was attempted to be made as robust as possible, but obviously still requires some manual changes depending on the specific dataset. The goal was to minimize the required changes, while still working properly and efficiently.

The dataset_creator.py contains two example scripts that were tested and seem to be working properly.
The first script is used for the brazil dataset, the second script (below the first, currently in triple brackets) is used for the mozambique2021 dataset.

The source for major parts of the above file was the notebook created by @noa-weiss and me for the mozambique2021 dataset, where I attempted to mimic Noa's great work and mold it into a robust and easy to use setting so that we can stay within the IDE and run everything with a click of a button the next time a new dataset arrives :)

### DIFF
--- a/soundbay/dataset_creator.py
+++ b/soundbay/dataset_creator.py
@@ -1,0 +1,335 @@
+import pandas as pd
+import os
+from sklearn.model_selection import train_test_split
+import numpy as np
+from soundbay.utils.metadata_processing import load_n_adapt_raven_annotation_table_to_dv_dataset_requirements
+
+
+class DatasetCreator:
+    def __init__(self, annotations_dir, dataset_name, save_dir, cols_to_use=None, label_col_name='label',
+                 desired_calls_label='w'):
+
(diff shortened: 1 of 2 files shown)


rafaelsul123:
Fixed the issue regarding the commit author, closed the previous PR.

Went over the script with Amit, seems to be working for his dataset as well. 

Further testing and comparison should be made between previous methods of generating datasets, to this current one - to make sure the desired functionallity is preserved.

amitgalor18 (approved):
Good PR. As discussed in our session - The script works well, as long as the user notices the annotation file names and the class labels to be included. After our fix it seems to be robust to different raven outputs

PR #73, opened 2023-06-28
Update ci_action.yml

Z30G0D:


rafaelsul123 (approved):
We found the source of error within a version update of audiomentations (0.30 -> 0.31) which caused the ci-test to break while building soxr-wheel.
Tests run through when using a specific version of audiomentations (==0.26), instead of greater than (>=0.26).

PR #75, opened 2023-07-11
bugfix/correct_start_times_on_inference

mosheman5:
bugfix: round the range of the arange, floating point issues might cause extra start_times and create short segments on inference

### DIFF
--- a/soundbay/data.py
+++ b/soundbay/data.py
@@ -3,6 +3,7 @@ from itertools import starmap, repeat
 from pathlib import Path
 from copy import deepcopy
 from typing import Union
+from decimal import Decimal
 
 import librosa
 import numpy as np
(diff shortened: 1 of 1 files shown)


soundbay/data.py
```diff
@@ -489,8 +492,10 @@ def _get_audio(self, begin_time):
         output:
         audio - pytorch tensor (1-D array)
         """
-        data, orig_sample_rate = sf.read(self.file_path, start=begin_time,
-                          stop=begin_time + int(self.seq_length * self.data_sample_rate))
+        duration = sf.info(self.file_path).duration
+        stop_time = begin_time + int(self.seq_length * self.data_sample_rate)
+        assert duration >= stop_time, f"trying to load audio from {begin_time} to {stop_time} but audio is only {duration} long"
```
Z30G0D, line 497: Should this be assert or just force the `stop_time` be equal to `duration`?

Z30G0D (approved):
LGTM!
Small question.

PR #81, opened 2023-09-19
Feature/paper touchups

mosheman5:
some touchups needed for the paper:
- Added explicit confs for all the 9 runs
- A script for data samples count
- A test csv for the orcasound dataset
- bugfix: use slice flag in inference
- add default path to the ICML2013 dataset


Z30G0D:
@mosheman5 
Did you check the CI after merge?

PR #83, opened 2023-11-15
Feature/new 2d models

mosheman5:
Add two new models, using pretrained versions that should improve performance on small datasets.
also add a preprocessor file that includes the melspectrogram representation, that should provide better results than Spectrogram (at least for some call types)

### DIFF
--- a/soundbay/conf/preprocessors/_mel_preprocessors.yaml
+++ b/soundbay/conf/preprocessors/_mel_preprocessors.yaml
@@ -1,0 +1,16 @@
+# @package _global_
+_preprocessors:
+  mel_spectrogram:
+    _target_: torchaudio.transforms.MelSpectrogram
+    sample_rate: ${data.sample_rate}
+    n_fft: ${data.n_fft}
+    hop_length: ${data.hop_length}
+    win_length: ${data.n_fft}
+    pad_mode: 'constant'
+    n_mels: 64
+    f_min: ${data.min_freq_filtering}
+    f_max: ${data.max_freq}
+  amplitude_2_db:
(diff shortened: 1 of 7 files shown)


soundbay/conf/preprocessors/_mel_preprocessors.yaml
shaing10, line 1: win_length = n_fft by definition - is that what we want?

BTW - https://github.com/deep-voice/soundbay/blob/brazil-slowfast/soundbay/conf/preprocessors/_preprocessors_mel_unit.yaml

soundbay/models.py
```diff
@@ -320,3 +320,47 @@ def forward(self, x):
...
+
+
+class Squeezenet2D(nn.Module):
+    def __init__(self, num_classes=2, pretrained=True):
+        super(Squeezenet2D, self).__init__()
+
+        self.squeezenet = torch.hub.load('pytorch/vision:v0.10.0', 'squeezenet1_0',
+                                         pretrained=pretrained)
+        # number of features from existing squeezenet
+        num_features = 1000
```
shaing10, line 332: if it is possible, maybe it is better to read it from `self.squeezenet` / from the config

shaing10 (approved):
2 small nice-to-have suggestions

soundbay/conf/preprocessors/_mel_preprocessors.yaml
mosheman5: https://pytorch.org/audio/main/generated/torchaudio.transforms.MelSpectrogram.html
It is the defaults behavior of melspecs, so I think it's gonna be fine.
regarding the BTW - lol. Any reason we didn't merge to master?

PR #88, opened 2024-02-18
Dataset/porpoise demo

mosheman5:


mosheman5:
- changed wrong commit author
- pushed porpoise data annotations
- small bugfix to handle correct max and min frequencies for augmentations and mel_scale representation

PR #90, opened 2024-03-06
Enhacement/unzip to s3

shaing10:
[thread-in-slack](https://deepvoicerd.slack.com/archives/C01633EV4HW/p1708183817499759)


shaing10:
After a discussion, thanks to @amitgalor18 's and @moshedeep 's notes, we have decided to treat txt file differently to make sure they are unzipped before uploading, but leave the other files treated as before - which is the more efficient way to deal with big files.

mosheman5 (approved):
no more gzip :)

PR #91, opened 2024-03-18
add path hierarchy, and support no bg dataset inference and training

mosheman5:


soundbay/data.py
```diff
@@ -36,11 +36,15 @@ def __init__(self, data_path, metadata_path, augmentations, augmentations_p, pre
         augmentations - list of classes audiogemtations
         augmentations_p - array of probabilities (float64)
         preprocessors - list of classes from preprocessors (TBD function)
+        path_hierarchy - enables working with data that is organized in a hierarchy of folders. The default value is 0,
+        which means all the audio files are flattened in the same folder. If the value is 1, the audio files are
```
Z30G0D, line 40: Understood it all,
but better just to give an example structure IMO.

soundbay/data.py
```diff
@@ -36,11 +36,15 @@ def __init__(self, data_path, metadata_path, augmentations, augmentations_p, pre
         augmentations - list of classes audiogemtations
         augmentations_p - array of probabilities (float64)
         preprocessors - list of classes from preprocessors (TBD function)
+        path_hierarchy - enables working with data that is organized in a hierarchy of folders. The default value is 0,
```
Z30G0D, line 39: The path_hierarchy variable should be an `Enum` IMO.
We will have an immutable set of options (integer) that have a semantic meaning:
```
from enum import Enum
class PathHierarchy(Enum):
      flat = 1
      per_class = 2
```      
And output of:
` PathHierarchy['per_class'].value`
will be `2`.

Z30G0D (changes requested):
I see you still handling path issues with Hydra.
Looks great!

soundbay/data.py
mosheman5: I think it's a good idea, but I don't think I have good names for hierarchies greater than 1. Also, it might be mis-leading since the nesting doesn't necessaraly be by the class. For example, I might wanna merge few data sources that have the save classes, but keep them in different folders

Z30G0D (approved):

shaing10:
default config does not run with this change. I'll fix it in #97

PR #92, opened 2024-04-27
Feature/tta

shaing10:
Test Time Augmentation - enables multi channels inputs where the channel dim contains spectrograms each with different set of parameters


shaing10:
With ICML data this feature does not seem to improve the result. The opposite, the levels of FP and FN have been slightly higher in this mode. But, the hyper parameters weren't optimized, and perhaps this data is not indicative enough. Tests would be conducted with more indicative data, where 2 spectrograms indeed reveal new info

PR #97, opened 2024-07-07
Feature/log time sample

shaing10:


soundbay/utils/logging.py
```diff
@@ -180,7 +181,7 @@ def upload_artifacts(self, audio: torch.Tensor, label: torch.Tensor, raw_wav: to
             specs.append(librosa.display.specshow(artifact_spec[artifact_id,...], ax=ax[1]))
             plt.close('all')
             del ax
-        list_of_specs_objects = [wandb.Image(data_or_path=spec, caption=f'label_{lab}_{ind}_train') for spec, ind, lab in zip(specs,idx, label)]
+        list_of_specs_objects = [wandb.Image(data_or_path=spec, caption=f'label{lab}_i{ind}_{round(b_t/sample_rate,2)}sec_{f_n}train') for spec, ind, lab, b_t, f_n in zip(specs,idx, label, meta['begin_time'], meta['org_file'])]
```
amitgalor18, line 184: Excellent idea to show the time in seconds

amitgalor18 (commented):
Great PR. No comments, looks good and makes the media section in w&b much more usable.

amitgalor18 (approved):

shaing10:
@amitgalor18 @moshedeep @Z30G0D 
I think there is a small bug here - the begin time in the description of the artifact is calculated as `begin_time/sample_rate`, in order to change the units into seconds. However, dividing by the sample rate is not correct, I should have divided by the original sample rate (data_sample_rate), am I right?

Z30G0D:
> @amitgalor18 @moshedeep @Z30G0D I think there is a small bug here - the begin time in the description of the artifact is calculated as `begin_time/sample_rate`, in order to change the units into seconds. However, dividing by the sample rate is not correct, I should have divided by the original sample rate (data_sample_rate), am I right?

That sounds about right.

PR #103, opened 2024-10-22
updated InferenceDataset to allow overlapping predictions.

danielle-hausler:
Added overlap argument, so that predictions can be done in overlapping windows.

### DIFF
--- a/soundbay/data.py
+++ b/soundbay/data.py
@@ -489,7 +489,8 @@ class InferenceDataset(Dataset):
                  preprocessors: DictConfig,
                  seq_length: float = 1,
                  data_sample_rate: int = 44100,
-                 sample_rate: int = 44100):
+                 sample_rate: int = 44100,
+                 overlap: float = 0):
         """
         __init__ method initiates InferenceDataset instance:
         Input:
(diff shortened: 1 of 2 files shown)


soundbay/data.py
```diff
@@ -540,8 +544,14 @@ def _create_start_times(self, filepath: Path) -> np.ndarray:
             audio_dict contains references to audio paths given name from metadata
         """
         audio_len = sf.info(filepath).duration
+        step = self.seq_length * (1-self.overlap)
         decimal_place = abs(Decimal(str(self.seq_length)).as_tuple().exponent)
```
shaing10, line 548: shouldn't the `decimal_place` be calculated with respect to the `step` instead of `self.seq_length` now that the starttimes vector is a function of `step`?
not sure what edge case it should catch to begin with actually but seems like it should be changed too here.

shaing10 (commented):
looks great.
maybe it's a different PR scope, but do we want to address this change also in `infer_without_metadata` function? post process the overlapping predictions into single prediction per time stamp? (majority vote or something smarter)
and maybe also pass this parameter from the config that deals with inference w/o metadata (the one that uses `inferenceDataset`)

soundbay/data.py
danielle-hausler: you are right, but actually i think i should delete the decimal_place and rounding, because its role is to handle the last element, however, since `step` could be shorter than` self.seq_length` (i.e., the range can end after `audio_len` - `self.seq_length`) i need to handle the last element anyway (with the filtering and the addition of the last start time). sorry for only noticing this now.

danielle-hausler:
@shaing10  thanks for checking this PR !!! 
updating inference_file.yaml is a good idea :) 
regarding `infer_without_metadata`, currently i'm using a more complex logic for my use case (removing adjacent windows where probability goes down by 10% and then merging the other windows), however, my use case is very specific and i'm not sure that i have something that could be used for all cases. maybe we can add in a different PR a set of functions (outside infer_without_metadata) that could be used for the processing of overlapping predictions?

shaing10 (approved):

PR #105, opened 2024-11-14
Feature/add efficientnet-b7 model

DorBernsohn:
Adding EfficientNet-B7 to enhance model options.

### DIFF
--- a/soundbay/models.py
+++ b/soundbay/models.py
@@ -369,19 +369,24 @@ class EfficientNet2D(nn.Module):
     """
     EfficientNet model for 3 channel ("RGB") input.
     """
-    def __init__(self, num_classes=2, pretrained=True):
+
+    def __init__(self, num_classes=2, pretrained=True, dropout=0.5, hidden_dim=256):
         super(EfficientNet2D, self).__init__()
-        
+
         # Load a pre-trained EfficientNet-B7 model from torchvision
-        self.efficientnet = models.efficientnet_b7(weights=EfficientNet_B7_Weights.DEFAULT) if pretrained else models.efficientnet_b7(weights=None)
-        
+        self.efficientnet = (
+            models.efficientnet_b7(weights=EfficientNet_B7_Weights.DEFAULT)
+            if pretrained
+            else models.efficientnet_b7(weights=None)
+        )
+
         # Replace the classification head to output the desired number of classes
         in_features = self.efficientnet.classifier[1].in_features
         self.efficientnet.classifier = nn.Sequential(
-            nn.Linear(in_features, 256),
+            nn.Linear(in_features, hidden_dim),
             nn.ReLU(),
-            nn.Dropout(0.5),
-            nn.Linear(256, num_classes)
+            nn.Dropout(dropout),
+            nn.Linear(hidden_dim, num_classes),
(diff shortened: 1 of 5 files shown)


soundbay/models.py
```diff
@@ -364,3 +364,26 @@ def __init__(self, num_classes=2, pretrained=True):
...
+
+class EfficientNet2D(nn.Module):
+    """
+    EfficientNet model for 3 channel ("RGB") input.
+    """
+    def __init__(self, num_classes=2, pretrained=True):
+        super(EfficientNet2D, self).__init__()
+        
+        # Load a pre-trained EfficientNet-B7 model from torchvision
+        self.efficientnet = models.efficientnet_b7(pretrained=pretrained)
```
shaing10, line 376: maybe we should have the chosen size `b7` as a parameter, coming from the configuration - I think that even `b0` worth trying and might be enough

soundbay/models.py
```diff
@@ -364,3 +364,26 @@ def __init__(self, num_classes=2, pretrained=True):
...
+        in_features = self.efficientnet.classifier[1].in_features
+        self.efficientnet.classifier = nn.Sequential(
+            nn.Linear(in_features, 256),
+            nn.ReLU(),
+            nn.Dropout(0.5),
+            nn.Linear(256, num_classes)
+        )
+
+    def forward(self, x):
+        x = x.repeat(1, 3, 1, 1)  # Repeat channel to convert 1-channel to 3-channel input
```
shaing10, line 388: suggestion: I think that @amitgalor18  once checked the effect of addressing spectrogram as an RGB image for real, I cannot remember the results or takeaways from this experiment. maybe it can be another adapter here, instead of repeating (?)

soundbay/models.py
```diff
@@ -364,3 +364,26 @@ def __init__(self, num_classes=2, pretrained=True):
...
+    """
+    EfficientNet model for 3 channel ("RGB") input.
+    """
+    def __init__(self, num_classes=2, pretrained=True):
+        super(EfficientNet2D, self).__init__()
+        
+        # Load a pre-trained EfficientNet-B7 model from torchvision
+        self.efficientnet = models.efficientnet_b7(pretrained=pretrained)
+        
+        # Replace the classification head to output the desired number of classes
```
shaing10, line 378: If we prefer not to explicitly change the classifier "from the outside", I think `EfficientNet` class supports this by just passing the argument
image

from: site-packages/torchvision/models/efficientnet.py

shaing10 (commented):
Looks great @DorBernsohn !
I left some comments inline if you'd like to consider, nothing too important. just my thoughts.
and, please add the model to `tests/test_models.py` such that it'd be tested as part of the CI

soundbay/models.py
Z30G0D, line 376: The `pretrained` parameter is deprecated.
It should be called like the resnet:
`resnet = models.resnet18(weights=ResNet18_Weights.DEFAULT) if pretrained else models.resnet18(weights=None)`

Z30G0D (changes requested):
LGTM,
small comment

shaing10 (approved):

PR #107, opened 2024-11-27
added wav2vec2 model, and related conf/utils files.

danielle-hausler:
main changes: 

1. added WAV2VEC2 to models.py (+updated conf_dict)
2. added files_handler to utils so that models for example from [https://github.com/earthspecies/aves] could be loaded. 
3. added custom conf files.

### DIFF
--- a/soundbay/conf/preprocessors/wav2vec2.yaml
+++ b/soundbay/conf/preprocessors/wav2vec2.yaml
@@ -1,0 +1,2 @@
+# @package _global_
+_preprocessors: []
--- a/soundbay/models.py
+++ b/soundbay/models.py
@@ -416,7 +416,7 @@ class EfficientNet2D(nn.Module):
         return self.efficientnet(x)
  
 
- class WAV2VEC2(nn.Module):
+class WAV2VEC2(nn.Module):
     def __init__(
             self,
             num_classes: int = 2,
--- a/soundbay/models.py
+++ b/soundbay/models.py
@@ -1,11 +1,17 @@
 import importlib
+from typing import Union
+
 import torch
+import torchaudio
 import torch.nn as nn
 from torch import Tensor
 from torchvision.models.resnet import ResNet, BasicBlock, conv3x3, Bottleneck
 from torchvision.models.vgg import VGG
 from torchvision.models import squeezenet, ResNet18_Weights
 import torchvision.models as models
+from torchaudio.models import wav2vec2_model
+
+from soundbay.utils.files_handler import load_config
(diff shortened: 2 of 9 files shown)


soundbay/conf/preprocessors/wav2vec2.yaml
```diff
@@ -0,0 +1,2 @@
+# @package _global_
+_preprocessors: []
```
Z30G0D, line 2: if there are no preprocessors why is this file needed and not use the `defaults`?

soundbay/models.py
```diff
@@ -1,11 +1,17 @@
...
+
 import torch
+import torchaudio
 import torch.nn as nn
 from torch import Tensor
 from torchvision.models.resnet import ResNet, BasicBlock, conv3x3, Bottleneck
 from torchvision.models.vgg import VGG
 from torchvision.models import squeezenet, ResNet18_Weights
 import torchvision.models as models
+from torchaudio.models import wav2vec2_model
```
Z30G0D, line 12: This is not necassary if we `import torchaudio`

soundbay/models.py
```diff
@@ -364,3 +370,51 @@ def __init__(self, num_classes=2, pretrained=True):
...
+        return self.fc(x)
+
+    def extract_features(self, x):
+        # this is separated from forward to allow feature extraction.
+        # sometimes for a batch of samples our raw input is [batch, 1, time]
+        if len(x.shape) > 2:
+            x = torch.squeeze(x, dim=1)
+        x = self.wav2vec.extract_features(x)[0]
+        # mean pooling over the layers: [layers, batch, time, features] -> [batch, time, features]
+        x = torch.stack(x, dim=0).mean(dim=0)
```
Z30G0D, line 408:
```suggestion
        x = torch.stack(x, dim=0).mean(dim=(0,2))
```
And remove the bottom return `mean`

soundbay/models.py
danielle-hausler: yep you are right

soundbay/conf/preprocessors/wav2vec2.yaml
danielle-hausler: wav2vec is working on the raw input (the waveform) and the defaults file has a preprocess of the data into a spectrogram.

soundbay/models.py
danielle-hausler: nice :)

Z30G0D (approved):

PR #108, opened 2024-12-04
intial step of adding multitarget support - adjusted BaseDataset

danielle-hausler:
# Adding Multitarget Label Support (Part 1): BaseDataset Adjustments  
This PR introduces the initial changes required to support multitarget labels, focusing on updating BaseDataset. 

### 1. **BaseDataset**  
- **Added `_preprocess_target` method:**  
  Handles string targets representing integers or lists of integers and is invoked within `preprocess_metadata`.  
- **Added `_is_noise` method:**  
  Checks if a label value represents noise.  
- **Slightly modified `_slice_sequence`:**  to work with `np.array` label values.  
- **Added `_get_num_classes` method:**  
  Calculates the number of classes based on the target type:  
  - Integer: counts unique values.  
  - `np.array` (possibly with overlapping classes), the array size determines the number of classes.  
- **Added `_get_samples_weight` method:**  
  Calculates sample weights differently based on target type:  
  - Integer: Inverse of the current row's class count.  
  - `np.array`: Weight is based on the rarest class in the array or noise count if no class is present.  
  Note: I kept this method within the class instead of moving it out as discussed, but I’m open to revisiting this decision if needed.  

### 2. **ClassifierDataset**  
- Minor change: Replaced direct comparisons of targets to `0` with the new `_is_noise` method.  

## Notes  
- This PR is the first part and multi-target experiments cannot yet be run successfully. Upcoming PRs will address logging, saved metrics, inference, and potentially corresponding conf files.  
- The `_get_audio` method was not simplified in this PR to keep the scope focused. This can be addressed in a separate PR.

### DIFF
--- a/soundbay/data.py
+++ b/soundbay/data.py
@@ -1,3 +1,4 @@
+import ast
 import random
 from itertools import starmap, repeat
 from pathlib import Path
@@ -66,6 +67,7 @@ class BaseDataset(Dataset):
         self.data_sample_rate = data_sample_rate
         self.sampler = torchaudio.transforms.Resample(orig_freq=data_sample_rate, new_freq=sample_rate)
         self._preprocess_metadata(slice_flag)
+        self.label_type = self._evaluate_label_type()
         self.augmenter = self._set_augmentations(augmentations, augmentations_p)
         self.preprocessor = self.set_preprocessor(preprocessors)
         assert (0 <= margin_ratio) and (1 >= margin_ratio)
@@ -106,7 +108,7 @@ class BaseDataset(Dataset):
         Output:
             ClassifierDataset object with self.metadata dataframe after applying the condition
         """
-        self._preprocess_target()
+        self.metadata['label'] = self._preprocess_target()
         is_noise = self.metadata['label'].apply(self._is_noise)
 
         # All calls are worthy (because we can later create a bigger slice contain them that is still a call in
@@ -125,7 +127,7 @@ class BaseDataset(Dataset):
 
         self.metadata.reset_index(drop=True, inplace=True)
 
-    def _preprocess_target(self):
+    def _preprocess_target(self) -> pd.Series:
         """
         Preprocesses the label column in the metadata. If the label is a string, it is evaluated and converted to an
         integer or a list of integers.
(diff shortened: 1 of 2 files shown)


soundbay/data.py
```diff
@@ -128,6 +125,26 @@ def _preprocess_metadata(self, slice_flag=False):
...
+        """
+        if pd.api.types.is_string_dtype(self.metadata['label']):
+            assert self.metadata['label'].str.match(r'^(\[|\()?(\d+)(\s*,\s*\d+)*(\]|\))?$').all(), \
+                "label should be a string that could be evaluated as a list of integers or integers."
+            self.metadata['label'] = self.metadata['label'].apply(eval)
+            if self.metadata['label'].apply(lambda x: isinstance(x, (list, tuple))).all():
+                self.metadata['label'] = self.metadata['label'].apply(np.array, dtype=int)
+
+    @staticmethod
+    def _is_noise(value: [int, np.ndarray]) -> bool:
```
shaing10, line 141: value: Union[int, np.ndarray] ?

soundbay/data.py
```diff
@@ -215,6 +234,34 @@ def set_preprocessor(preprocessors_args):
             preprocessor = torch.nn.Identity()
         return preprocessor
 
+    def _get_num_classes(self) -> int:
+        """
+        Returns the number of classes in the metadata.
+        """
+        if self.metadata['label'].apply(lambda x: isinstance(x, np.ndarray)).all():
```
shaing10, line 241: what does this condition check? when will we get here? is it guaranteed that ndarrays are the representations of the multilabel case?

shaing10 (commented):
Hey, LGTM! A couple of small comments:

Regarding inputs and outputs of methods versus modifying attributes within the method's scope - see my comment inline: this is more of a styling preference, so feel free to ignore, but I find explicit inputs/outputs often improve readability.

About identifying the different modes based on the type (e.g., ndarray vs. int): I’m not sure this approach is explicit or readable enough. More importantly, it seems to couple the implementation from the algorithm. If possible, could the condition rely on a more descriptive or domain-specific distinction?

Thanks!

soundbay/data.py
```diff
@@ -108,13 +106,12 @@ def _preprocess_metadata(self, slice_flag=False):
         Output:
             ClassifierDataset object with self.metadata dataframe after applying the condition
         """
+        self._preprocess_target()
```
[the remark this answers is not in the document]
danielle-hausler: So regarding the different modes comment - I've added a `label_type` attribute and an `_evaluate_label_type` method to identify the different modes. I think this approach improves clarity, but let me know if there's still room for improvement.

Regarding the comment on explicit inputs/outputs: I understand your point and gave it some thought. However, I'm not sure about passing an attribute as an input, It feels a bit awkward. The `label_type` attribute that I added specifically refers to the label column in the metadata, and I think this should remain the default input for most of these functions. Allowing a more general input might dilute this focus, so I’ve avoided it for now.

Do you have any suggestions for making the implementation more readable without passing class attributes as inputs? I'm open to ideas :)

soundbay/data.py
danielle-hausler: I've modified this part, but basically after the target preprocessing ndarrays should be the representations of the multi-label case.

shaing10 (approved):

PR #109, opened 2025-01-15
modified logging and trainers to for multi-target

danielle-hausler:
Apologies for the large number of file changes in this PR! :( 

To expedite progress on the multi-label feature, I’m pushing this branch, which is still a work in progress. However, the training flow runs without errors, and this should give us a good starting point for review.

Note that some files have minimal changes: 
* config files that were changed (label_type parameter was added) 
-> `soundbay/conf/data/defaults.yaml` 
-> `soundbay/conf/data/multi_label.yaml`  
We might want to set a default value to the label_type for backward compatibility (so we might need a small change in the data.py file) 

* BCEloss was added to: 
-> `soundbay/conf_dict.py` 
-> soundbay/train.py

More important changes: 
-> in `soundbay/data.py` instead of evaluating the label column we have a config parameter that sets it's type. 
-> the metrics calculation in the logger were moved to `soundbay/utils/metrics.py` 
-> the probabilities post processing calculation was moved to  `soundbay/utils/post_process.py` 

Here the next things that should be done: 
### small things: 
1.~~add to `soundbay/utils/conf_validator.py` a validation of the type of the label column and make sure it matches `label_type`.~~ on a second thought we have assertions inside the data class that already do this. 
2. ~~maybe give a default value to `label_type`.~~
3. ~~allow passing the `th` argument in `post_process_predictions` from external inputs.~~
4.~~enable WandB curves for multi-target scenarios (currently disabled).~~ unfortunately WandB curve functions do not support this (the shape of the target value should be N, while the shape of the probabilities array should be (N, C). 
5. ~~modify inference: `predict_proba` should use sigmoid for label_type='multi_label' and softmax for 'single_label'.~~
6.~~add a model conf file with BCEloss for multi target.~~

### more work consuming changes: 
7. dataset creation: Write a function to handle overlapping time intervals.
given a sequence of intervals [i1=[t0,t1], i2=[t2,t3], i3=[t4,t5]], return for each interval the elements it overlaps with, e.g.:
[[i1], [i2,i3], [i2,i3]] or [[1,0,0], [0,1,1], [0,1,1]].
8. Tests

### Update 26.1 
- for backward compatibility i used .get for the new config args with the previous value as default value. 
- the additional files that were added in the last commit are the inference files.


soundbay/trainers.py
```diff
@@ -113,18 +118,23 @@ def train_epoch(self, epoch):
 
             # estimate and calc losses
             estimated_label = self.model(audio)
+            if self.label_type == 'multi_label':
```
shaing10, line 121: why the if statement? doesn't this `as_type` fit and can benefit all modes?

soundbay/trainers.py
```diff
@@ -113,18 +118,23 @@ def train_epoch(self, epoch):
...
+                label = label.type_as(estimated_label)
             loss = self.criterion(estimated_label, label)
             loss.backward()
             self.optimizer.step()
 
+            # process the logit predictions:
+            predicted_proba, predicted_label = post_process_predictions(estimated_label.data, self.label_type, self.proba_threshold)
```
shaing10, line 128: does `estimated_label` always have `data` attribute? maybe this line (128) should be wrapped with if statement for multilabel mode only?

soundbay/trainers.py
```diff
@@ -152,15 +162,20 @@ def eval_epoch(self, epoch: int, datatset_name: str = None):
 
                 # estimate and calc losses
                 estimated_label = self.model(audio)
+                if self.label_type == 'multi_label':
+                    label = label.type_as(estimated_label)
                 loss = self.criterion(estimated_label, label)
 
+                # process the logit predictions:
+                predicted_proba, predicted_label = post_process_predictions(estimated_label.data, self.label_type, self.proba_threshold)
```
shaing10, line 170: same as previous comment

soundbay/utils/post_process.py
```diff
@@ -0,0 +1,18 @@
+import torch
+
+
+def post_process_predictions(preds: torch.Tensor, label_type: str, th: float = None) -> tuple:
+    """
+    Post-process the predictions to probabilities
+    """
+    if th is None:
```
shaing10, line 8: non-significant-comment - you can set the default value in the signature as 0.5 and remove lines 8-9 if you'd like, i think it'll be the same

soundbay/utils/logging.py
```diff
@@ -135,7 +140,7 @@ def calc_metrics(self, epoch: int, mode: str = 'train', label_names: Optional[Li
                 self.log_writer.log({f'Call Metrics {mode}/{metric}_{label_names[class_id]}':
                                     self.metrics_dict['calls'][class_id][metric]}, step=epoch)
 
-        if not self.debug_mode:
+        if (not self.debug_mode) and isinstance(self.label_list[0], int) == 1:
```
shaing10, line 143: can you explicitly define what this condition means?
```
cond-with-meaningful-name = isinstance(self.label_list[0], int) == 1 # is the "==1" redundant?
if (not self.debug_mode) and cond-with-meaningful-name: 
...
```
if i get it right, `label_list` can be list of ints or list of numpy arrays, for mode "single" and mode "multi" respectively, and here we basically verify the mode?   if it's the case, maybe it'll be clearer to pass the label_mode explicitly

soundbay/utils/metrics.py
shaing10, line 1: really like this change! I think the metrics deserve a class, and not just a function in the logger. but, since it's a big change, i wonder how we can verify it. the previous code used to update attributes over time, and with the new design we instantiate the `MetricsCalculator` everytime w/o sharing info between instances. does it work the same as before? how?

shaing10 (commented):
loved it! left some comments inline.
I haven't approved yet since i know you small new changes you'd like to add

soundbay/conf/data/defaults_inference.yaml
```diff
@@ -3,6 +3,7 @@ data:
   batch_size: 64
   num_workers: 10
   data_sample_rate: 44100
+  proba_norm_func: 'softmax'
```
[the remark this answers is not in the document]
danielle-hausler: good idea :) since the training config contains the argument 'label_type' i've set the `proba_norm_func` according to `ckpt_args.data.label_type`

soundbay/inference.py
```diff
@@ -49,8 +52,13 @@ def predict_proba(model: torch.nn.Module, data_loader: DataLoader,
                     return predicted_probability[:, selected_class_idx]
                 else:
                     raise ValueError(f'selected class index {selected_class_idx} not in output dimensions')
-        softmax_activation = softmax(all_predictions, 1)
-        return softmax_activation
+        if proba_norm_func == 'softmax':
+            proba = softmax(all_predictions, 1)
+        elif proba_norm_func == 'sigmoid':
+            proba = expit(all_predictions)
```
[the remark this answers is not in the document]
danielle-hausler: since we use softmax from scipy.special, i used the same package.

soundbay/conf/data/defaults_inference.yaml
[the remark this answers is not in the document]
danielle-hausler: wait, i'm not sure this is a good idea
i think we should allow some flexibility and let the user choose the normalization function, otherwise my previous trained models will not work since they are old and their config do not contain "label_type".

soundbay/trainers.py
danielle-hausler: nope unfortunately this is an annoying issue with torch types and the loss functions..

soundbay/utils/metrics.py
danielle-hausler: `MetricsCalculator` is a new class for metrics calculation, and the logger instantiate the `MetricsCalculator` and updates and logs the metrics over time, so it works as before :)

soundbay/conf/data/defaults_inference.yaml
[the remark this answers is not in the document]
shaing10: I think that models that have trained with one norm function would not perform well with different one... but, I agree that backward compatibility makes it more problematic in the transition time.
I suggest to have the pointer pushed to the default config, and each user will still have the option to overwrite it if s/he wants. IMO it's better to get an error that will make you aware of a mismatch or missing param, than to silently inference with the wrong function and try to understand why the results are worse than what you've seen during training time...
but I see the drawbacks too. I count on your judgement, whatever you decide

shaing10 (approved):
@danielle-hausler you are the best! I left some additional comments regarding the new function, but they are all nice-to-have.
I felt like having a deja-vu though; haven't you pushed something similar in the past? a few months ago? not in the context of multilabel, but in the context of identifying and processing overlapping. if we have, maybe it's worth considering to extract the overlapping logic into a standalone code (function i guess) if possible and reuse it in both places.

Thanks!

soundbay/utils/metadata_processing.py
```diff
@@ -254,3 +255,55 @@ def bg_from_non_overlap_calls(df: pd.DataFrame):
...
+        1    [0, 1]
+        2    [1, 0]
+    """
+    Interval = namedtuple('Interval', ['start', 'end', 'label'])
+    n_classes = df["label"].nunique() - (noise_class_value in df["label"].unique())
+    overlaps = {idx: [0] * n_classes for idx in df.index}
+    reference_intervals = {}
+
+    # Process intervals in chronological order
+    for idx, row in df.query('label != 0').sort_values('begin_time').iterrows():
```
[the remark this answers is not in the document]
danielle-hausler: oops, the 0 should not be there i'll fix that !

soundbay/utils/metadata_processing.py
```diff
@@ -254,3 +255,55 @@ def bg_from_non_overlap_calls(df: pd.DataFrame):
...
+        >>> start_times = np.random.uniform(0, 10, 3)
+        >>> end_times = start_times + 1
+        >>> labels = np.random.choice([1,2], 3)
+        >>> df = pd.DataFrame({'begin_time': start_times, 'end_time': end_times, 'label': labels})
+        >>> df
+                begin_time	end_time	label
+            0	4.051811	5.051811	2
+            1	8.789995	9.789995	2
+            2	5.861857	6.861857	1
+    >>> multi_target_from_time_intervals_df(df, min_overlap_threshold=0.5, noise_class_value=0)
```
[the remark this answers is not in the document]
danielle-hausler: yep its just a quick example i created to show how the output looks like, but i can modify one of the values to make it overlap :)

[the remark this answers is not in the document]
danielle-hausler: regarding the overlapping logic, you are right i should extract it into a separate function.

soundbay/conf/data/defaults_inference.yaml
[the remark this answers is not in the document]
danielle-hausler: i think i can replace` args.data.get('proba_norm_func', 'softmax')` with a default value that is based on `label_type`.

soundbay/utils/metadata_processing.py
```diff
@@ -254,3 +255,55 @@ def bg_from_non_overlap_calls(df: pd.DataFrame):
     bg_df = pd.concat(bg_calls)
     bg_df['label'] = 0
     return pd.concat([bg_df, df], ignore_index=True)
+
+
+def multi_target_from_time_intervals_df(df, min_overlap_threshold: float = 0.0, noise_class_value: int = 0) -> pd.Series:
```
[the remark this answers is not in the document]
shaing10: so maybe we should publish your notebook or a simpler version of it, to have an example of how to run with this new mode. in the future :)

soundbay/utils/metadata_processing.py
```diff
@@ -273,21 +306,24 @@ def multi_target_from_time_intervals_df(df, min_overlap_threshold: float = 0.0,
         >>> df = pd.DataFrame({'begin_time': start_times, 'end_time': end_times, 'label': labels})
         >>> df
                 begin_time	end_time	label
-            0	4.051811	5.051811	2
+            0	4.051811	6.051811	2
             1	8.789995	9.789995	2
             2	5.861857	6.861857	1
-    >>> multi_target_from_time_intervals_df(df, min_overlap_threshold=0.5, noise_class_value=0)
-        0    [0, 1]
+    >>> multi_target_from_time_intervals_df(df, overlap_threshold_pct=0.5, noise_class_value=0)
```
shaing10, line 312: i thought I understood but this example now confuses me: the overlapping between [4.05...-6.05...] and [5.8..-6.8..] is 0.25  absolutely and ~0.2 relatively which is lower than the th 0.5, so why both labels present for the section?

danielle-hausler: sorry it should be overlap_threshold_pct=0

PR #114, opened 2025-03-31
Feature/ssl ast perf test

shakediel:
this PR contains the changes to run different comparison between AST model to resnet182D as detailed [here](https://app.clickup.com/t/86c22868b)

main changed include:
* adding AST preprocessor to configs
* add new scheduler to configs
* updating mozambique dataset config to support new format
* update AST to load local weights from path


shakediel:
Experiments results can be found [here](https://wandb.ai/deep_voice/ast_ssl_mozambique_2018?nw=nwuserdeep_voice)

soundbay/conf/data/mozambique_2018.yaml
```diff
@@ -1,19 +1,22 @@
...
   min_freq_filtering: 0
   n_fft: 1024
   hop_length: 256
+  label_type: 'single_label'
   train_dataset:
     _target_: soundbay.data.ClassifierDataset
     data_path: '../datasets/mozambique_2018'
+    path_hierarchy: 0
     mode: train 
-    metadata_path: '../datasets/mozambique_2018/combined_annotations_filtered_train.csv'
```
shaing10, line 16: it does not work for me w/o the relative path...

soundbay/conf/optim/defaults.yaml
```diff
@@ -3,8 +3,9 @@ optim:
...
   optimizer:
     _target_: torch.optim.Adam
-    lr: 0.001
+    lr: 5e-4
   scheduler:
-    _target_: torch.optim.lr_scheduler.ExponentialLR
-    gamma: 0.995
-  freeze_layers_for_finetune: False # Default is False.
\ No newline at end of file
+    _target_: torch.optim.lr_scheduler.CosineAnnealingWarmRestarts
```
shaing10, line 8: should we adopt "CosineAnnealingWarmRestarts" IYO?

soundbay/conf/optim/defaults.yaml
```diff
@@ -3,8 +3,9 @@ optim:
...
+    kwargs:
+      T_0: 5
+  freeze_layers_for_finetune: True # Default is False.
```
shaing10, line 11: default should be False I guess

soundbay/models.py
```diff
@@ -49,33 +48,47 @@ def freeze_layers(self):
...
+        return mel_spectogram['input_values'].squeeze()
+
 class AST(nn.Module):
     def __init__(self, weight_path, num_classes):
         super(AST, self).__init__()
-        self.processor = AutoProcessor.from_pretrained("MIT/ast-finetuned-audioset-10-10-0.4593", max_length=1024)
 
         self.model = ASTModel.from_pretrained("MIT/ast-finetuned-audioset-10-10-0.4593")
-        state_dict = torch.load(weight_path)
+        if weight_path != 'None':
```
shaing10, line 67: instead of "None" as string, you can insert `null` in the yaml config, and then real `None` will be assigned to the variable

soundbay/models.py
```diff
@@ -49,33 +48,47 @@ def freeze_layers(self):
...
+        mel_spectogram = self.processor(orig_wav.numpy(), sampling_rate=self.sample_rate, return_tensors="pt")
+
+        return mel_spectogram['input_values'].squeeze()
+
 class AST(nn.Module):
     def __init__(self, weight_path, num_classes):
         super(AST, self).__init__()
-        self.processor = AutoProcessor.from_pretrained("MIT/ast-finetuned-audioset-10-10-0.4593", max_length=1024)
 
         self.model = ASTModel.from_pretrained("MIT/ast-finetuned-audioset-10-10-0.4593")
```
shaing10, line 66: better to be loaded from the config (2)?

soundbay/models.py
```diff
@@ -49,33 +48,47 @@ def freeze_layers(self):
             param.requires_grad = True
 
 
+
+class ASTPreprocessorWrapper():
+    def __init__(self, sample_rate):
+        self.processor = AutoProcessor.from_pretrained("MIT/ast-finetuned-audioset-10-10-0.4593", max_length=1024)
```
shaing10, line 54: better to be loaded from the config?

soundbay/models.py
```diff
@@ -49,33 +48,47 @@ def freeze_layers(self):
...
 
         self.fc = nn.Linear(self.model.config.hidden_size, self.num_classes)
 
     def forward(self, x):
-        # Assuming x is the input that needs to be processed before passing to the model
-        import ipdb;
-        sampling_rate = 16000 #TODO: add sample rate from config --> only takes this!  
-        mel_spectogram = self.processor(x.squeeze(1).cpu().numpy(), sampling_rate = sampling_rate, return_tensors="pt")
-        
-        inputs = {key: val.to(x.device) for key, val in mel_spectogram.items()}  # Move to device
```
shaing10, line 71: don't you need it anymore?

split_csv_stratified.py
shaing10, line 1: do we need it in the code base?
if yes, maybe not in the root dir but in one of the folders

shaing10 (commented):
LGTM! `AST` is a really important contribution as model candidate :)
I left some comments inline.
I haven't delved into the details, but the code related to AST seemed very straight forward and simple to follow. please make sure that the transformers version you worked with is the one in the requirements file.

Moreover, please add this new model to the tests in `tests/test_models.py`, such that it'll be tested as part of the CI.

Thanks! Great first PR. Whalecome :)

soundbay/conf/optim/defaults.yaml
danielle-hausler: Looks like these parameters are intended for an optimizer that fine-tunes an AST model (or maybe for fine-tuning other models as well). Would it make sense to separate them into an additional config file specifically for that use case?

soundbay/models.py
danielle-hausler, line 54: Can we define a custom max_length variable here? What are the implications of using a different sampling rate than what the model was trained on?

soundbay/train.py
```diff
@@ -152,7 +167,7 @@ def modeling(
     optimizer_args = dict(optimizer_args)
     optimizer = optim_dict[optimizer_args.pop('_target_')](model.parameters(), **optimizer_args)
 
-    scheduler = scheduler_dict[scheduler_args._target_](optimizer, gamma=scheduler_args['gamma'])
+    scheduler = scheduler_dict[scheduler_args._target_](optimizer, **scheduler_args['kwargs'])
```
danielle-hausler, line 170: if we’re now expecting `scheduler_args['kwargs']`, could this break compatibility with previous runs or configs that didn’t use that structure?

soundbay/train.py
```diff
@@ -191,8 +206,13 @@ def main(validate_args) -> None:
         _logger = wandb
 
     experiment_name = get_experiment_name(args)
-    _logger.init(project="finding_willy", name=experiment_name, group=args.experiment.group_name,
-                 id=args.experiment.run_id, resume=args.experiment.checkpoint.resume)
+    _logger.init(
+        project=args.experiment.project,
```
danielle-hausler, line 210: Not sure if we currently have a 'project' variable in the experiment config, could this break previous runs that don’t include it? Maybe we should set a default value to ensure backward compatibility.
Also, I’d suggest running the main config file (not the AST specific one) to make sure everything still works as expected before merging.

danielle-hausler (commented):
Looks great! :)
I’m really excited that we’re finally adding AST to our codebase !!
I’ve just left a few minor comments below.

soundbay/conf/optim/defaults.yaml
shakediel: dont see a reason not to, it produced good results on the baseline (resnet18 + full dataset) and in general it is a good practice scheduler

soundbay/conf/optim/defaults.yaml
shakediel: Based on the experiments I ran it seems that freezing the layers produced a more stable training graph, converged faster and did not harm the results, see I suggest we keep the default True
any objections?

soundbay/models.py
shakediel: i dont think the max_length directly implies the sampling rate, according to the docs it seems this implies the max sample length, since its a transformer i assume it can be safely modified but i havent tested it
about the sample_rate, i assume we can approach this as a data drift and finetune the model sample_rate, its an interesting question that should be further investigated imo :)

soundbay/models.py
shakediel: nope, most of the code was moved to ASTPreprocessorWrapper

soundbay/train.py
shakediel: thanks! i updated the other configs to also include 'kwargs' as a wrapper to their parameters

soundbay/train.py
shakediel: i added a default parameter to the config that now holds 'finding_willy' in the default config

shakediel:
@shaing10 i added AST to the test_modles script but i see it fails due to missing parameters that are usually passed by the config that does not exist in the test
any suggestion on how to approach this? Should i just add a default value to the model parameters?

soundbay/train.py
mosheman5: not sure I understand what's going on with the other kwargs, I think it's kinda redundant and we can use `**scheduler_args`, instead of `**scheduler_args['kwargs']` no?

shaing10:
> @shaing10 i added AST to the test_modles script but i see it fails due to missing parameters that are usually passed by the config that does not exist in the test any suggestion on how to approach this? Should i just add a default value to the model parameters?

@shakediel so sorry, I missed this message. not sure I'm getting the problem. it makes sense that you'll need to update the test config. we can have a pair on this

mosheman5 (approved):

PR #119, opened 2025-08-06
Bug fixes

uvmend:
Fixed several bugs in the code:

(1)
**inference:** infer_with_metadata was broken as it used an unsupported function. I also changed in MetricsCalculator the __init__ function so there is no need to convert pred_proba from np.ndarray to list and back (can be done for the other lists, but I was not sure we want to)

(2)
**seq_length:** In config files, the seq_length has to be the same for training and validation.

(3)
**max_freq** removal:
This variable is set to be half the sample rate, so we need to have it in the config file. It created a series of removals of unused variables. Please double-check me that I did not remove any vital variable.

(4)
**updated .gitignore:**
We don't need to add the wandb files to the git by mistake.

(5)
**train in debug mode:**
This option was broken. Needed to initialize the Mock() class in debug mode as well (removed one condition).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,4 +34,8 @@ outputs/
 /soundbay/.null/inference.log
 /soundbay/wandb/
 
+# datasets
+datasets/fannie_project
+
+# wandb
 wandb/*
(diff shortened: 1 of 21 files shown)


.gitignore
```diff
@@ -34,4 +34,8 @@ outputs/
 /soundbay/.null/inference.log
 /soundbay/wandb/
 
+# datasets
+datasets/fannie_project
```
mosheman5, line 38: do we want this here?

uvmend: I thought it could be helpful to keep it as a way for other people to add their datasets when working, without unintentionally uploading them to the git.
But I can remove it as well. its not important.

PR #124, opened 2026-02-10
Adding devcontainer

Z30G0D:


mosheman5 (changes requested):
Looks good, I would suggest to add a small part in the README that explains how to use it

PR #125, opened 2026-04-14
Feature/raven export wrapper

shaing10:
## Summary

- Adds `scripts/raven_export.py` — exports soundbay checkpoints (ResNet182D, Squeezenet2D) to Raven Intelligence `.ravenmodel` packages with preprocessing baked in
- Supports TorchScript (recommended) and ONNX export formats
- Re-implements PeakNormalize, UnitNormalize, and LibrosaPcen as `nn.Module`s for TorchScript compatibility
- Adds `--compensate-for-sigmoid` flag (recommended): Raven applies sigmoid to model outputs internally, so this option outputs `inverse_sigmoid(softmax(logits))` — Raven's sigmoid then recovers exact softmax probabilities that sum to 1.0

## Usage

**Recommended** (until Raven publishes a new version with more DOFs in their post-processing logic): compensate for Raven's sigmoid post-processing

```bash
python scripts/raven_export.py /path/to/checkpoint/best.pth \
    --output-dir ~/Raven\ Workbench/Raven\ Intelligence/Models/ \
    --name My_Detector \
    --compensate-for-sigmoid
```

**Raw logits** (no activation, Raven applies sigmoid):

```bash
python scripts/raven_export.py /path/to/checkpoint/best.pth \
    --name My_Detector --no-softmax
```

## Why some modules are re-implemented

TorchScript export requires all operations to be `nn.Module` instances or pure torch ops. Several soundbay preprocessors can't be traced as-is:

| Module | Why re-implemented | Approach |
|---|---|---|
| `PeakNormalize` | Plain class with `__call__`, not `nn.Module` — TorchScript can't trace it | Thin `nn.Module` wrapper with identical logic (verified: 0.0 diff) |
| `UnitNormalize` | Same — not `nn.Module` | Same approach (verified: 0.0 diff) |
| `LibrosaPcen` | Calls `librosa.core.pcen` (NumPy/SciPy) — can't run in TorchScript | Full reimplementation in pure PyTorch with `torch.jit.script` for the IIR loop. Matches librosa's filter coefficients and `lfilter_zi` initialization exactly |
| `MinFreqFiltering` | Uses `numpy.floor` | Replaced with `nn.Identity` (validated: only `min_freq=0` configs are supported; errors on non-zero) |

## Test plan

- [x] Exported barks detector with all 3 modes (softmax, no-softmax, compensate-for-sigmoid)
- [x] Verified compensated model scores sum to exactly 1.0 after Raven's sigmoid
- [x] Confirmed detection agreement across all export modes on real WAV file
- [x] Tested in Raven Intelligence with TH=0.2

### DIFF
--- a/scripts/raven_export.py
+++ b/scripts/raven_export.py
@@ -4,6 +4,86 @@ Raven Intelligence Export Module
 This module provides utilities to export soundbay models to Raven Intelligence format.
 It wraps trained models with their preprocessing pipeline to create standalone models
 that accept raw audio input, compatible with Raven Intelligence's expected interface.
+
+Raven Intelligence is a bioacoustics analysis tool from Cornell Lab of Ornithology that
+uses DJL (Deep Java Library) to run ML models for automated sound detection.
+
+=========================================================================================
+USAGE EXAMPLES
+=========================================================================================
+
+1. COMMAND LINE - Export a soundbay checkpoint to Raven format:
+   -------------------------------------------------------------------------
+   python scripts/raven_export.py /path/to/checkpoint/best.pth \\
+       --output-dir ~/Raven\ Workbench/Raven\ Intelligence/Models/ \\
+       --name My_Detector
+
+   This requires args.yaml to be in the same directory as the checkpoint.
+
+2. PYTHON API - Export from soundbay checkpoint:
+   -------------------------------------------------------------------------
+   from scripts.raven_export import create_raven_model_package
+
+   create_raven_model_package(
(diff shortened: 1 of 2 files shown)


mosheman5 (commented):
looks good
do we still new --compensate-for-sigmoid in the default/recommended command?

shaing10:
> looks good do we still new --compensate-for-sigmoid in the default/recommended command?

image
seems like the new release solved it, I'll change the default

mosheman5 (approved):

